PR #12, opened 2021-01-07
login_client improvements

### LINKED ISSUE #11: Add Stream for Credentials changes
Right now it's quite hard to manage the circular dependency on some kind of authentication (say) cubit and the login client itself. The authentication thing needs the login client, but the login client needs to somehow "push" the credentials changes back to this authentication thing. Streams were initially used in the `leancode_login_client` and they served a good purpose for that. 

Right now it uses a simple callback because the OAuth client used it too, but for this purpose, a stream looks like a much more viable solution.

Albert221:
After using this library in a production app, there were few issues discovered. This Pull Request fixes them.

Also, I went back to the `onCredentialsChanged` stream in the `LoginClient` class as it was in the 0.* versions of the login_client library. It's just much easier for the developer and cleaner to use stream instead of passing a callback that the developer would have to prospectively somehow reach from other parts of the code.

**login_client**

- OAauth2 `clientId` and `clientSecret` now default to an empty string.
- Export `oauth2` `Credentials`, `AuthorizationException` and `ExpirationException`.
- Add `onCredentialsChanged` stream to `LoginClient`.
- Deprecate `credentialsChangedCallback`.

**login_client_flutter**

- Fix reading null credentials.

Closes #11.

### DIFF
--- a/packages/login_client/CHANGELOG.md
+++ b/packages/login_client/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 1.1.0
 
-- OAauth2 `clientId` and `clientSecret` now default to an empty string.
+- OAuth2 `clientId` and `clientSecret` now default to an empty string.
 - Export `oauth2` `Credentials`, `AuthorizationException` and `ExpirationException`.
 - Add `onCredentialsChanged` stream to `LoginClient`.
 - Deprecate `credentialsChangedCallback`.
--- a/packages/login_client/CHANGELOG.md
+++ b/packages/login_client/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 1.1.0
 
-- OAuth2 `clientId` and `clientSecret` now default to an empty string.
+- OAuth2 `clientId` is now required.
+- OAuth2 `clientSecret` now defaults to an empty string.
 - Export `oauth2` `Credentials`, `AuthorizationException` and `ExpirationException`.
 - Add `onCredentialsChanged` stream to `LoginClient`.
 - Deprecate `credentialsChangedCallback`.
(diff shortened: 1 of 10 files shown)


packages/login_client/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 1.1.0
+
+- OAauth2 `clientId` and `clientSecret` now default to an empty string.
```
jakubfijalkowski, line 3: Typo: `Aa`

packages/login_client/lib/src/oauth_settings.dart
```diff
@@ -19,10 +19,11 @@ class OAuthSettings {
   /// Creates the [OAuthSettings].
   const OAuthSettings({
     @required this.authorizationUri,
-    this.clientId,
-    this.clientSecret,
+    this.clientId = '',
```
jakubfijalkowski, line 22: `clientId` is required by the spec (and all implementations). Moreover, I don't think it can ever be empty - I haven't seen any OAuth2 servers accepting empty `clientId` (although, AFAIK, spec doesn't disallow that).

packages/login_client/lib/src/login_client.dart
```diff
@@ -62,28 +65,38 @@ class LoginClient extends http.BaseClient {
     @required OAuthSettings oAuthSettings,
     @required CredentialsStorage credentialsStorage,
     http.Client httpClient,
-    CredentialsChangedCallback credentialsChangedCallback,
+    @Deprecated('Use onCredentialsChanged instead.')
```
jakubfijalkowski, line 68: Do you have any deprecation plan?

Albert221: Look at #13.

mateuszwojtczak (approved):

PR #16, opened 2021-01-11
Fix for null query results

Albert221:
Related to https://github.com/leancodepl/corelibrary/issues/193.

### DIFF
--- a/packages/cqrs/lib/src/cqrs.dart
+++ b/packages/cqrs/lib/src/cqrs.dart
@@ -73,6 +73,11 @@ class CQRS {
       try {
         final json = jsonDecode(response.body);
 
+        // Fix for https://github.com/leancodepl/corelibrary/issues/193
+        if (json == null) {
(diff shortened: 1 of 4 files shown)


packages/cqrs/lib/src/cqrs.dart
```diff
@@ -73,6 +73,11 @@ class CQRS {
       try {
         final json = jsonDecode(response.body);
 
+        // Fix for https://github.com/leancodepl/corelibrary/issues/193
```
mchudy, line 76: I don't think we need this comment, the reason for this change is well-documented in Git and GH

Albert221: Actually, it's the contracts generator that causes this issue, this is only a slice to fix it. IMHO it's nice to leave it here because it's not our code that is actually responsible for the exception

mchudy (approved):

Issue #14, opened 2021-01-07
Remove deprecated success parameter from CommandResult


Albert221:
https://github.com/leancodepl/flutter_corelibrary/blob/bf67a60b5d4cffc9b1807445bc5b3569b72d2b5f/packages/cqrs/lib/src/command_result.dart#L21

PR #21, opened 2021-03-10
Add propertyName to ValidationError

mchudy:
I'm not sure what's the version we finally want to go with, as technically it's a breaking change I bumped it to 5.0.0.

### DIFF
--- a/packages/cqrs/CHANGELOG.md
+++ b/packages/cqrs/CHANGELOG.md
@@ -1,5 +1,6 @@
 # 5.0.0
 
+- Revert to 4.1.1
 - **Breaking:** Add `propertyName` field in `ValidationError`.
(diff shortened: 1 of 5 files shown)


packages/cqrs/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 5.0.0
+
+- **Breaking:** Add `propertyName` field in `ValidationError`.
+
```
Albert221, line 4: I would include the past versions (5.0.0 prereleases) here. They do exist on pub.dev so imo they should be here

Albert221 (changes requested):
I don't remember, will this 5.0.0 include the removal of a deprecated method (see #14)?

mchudy:
It doesn't, but I think we can remove it as it's a new major

Albert221 (approved):
LGTM!

PR #29, opened 2021-07-20
Add legacy changelog entry

mateuszwojtczak:


Albert221 (approved):

Albert221:
Don't we also want to publish the version as the 2.0.3+1?

mateuszwojtczak:
> Don't we also want to publish the version as the 2.0.3+1?

done! :)

PR #32, opened 2021-08-02
Add possibility to create login client credentials storage for web and mobile

FlowerAlex:


packages/login_client_flutter/lib/src/flutter_web_secure_credentials_storage.dart
```diff
@@ -0,0 +1,49 @@
...
+  @override
+  Future<Credentials?> read() async {
+    try {
+      final json = window.localStorage['credentials'];
+      if (json != null) {
+        return Credentials.fromJson(json);
+      } else {
+        return null;
+      }
+    } catch (e) {
```
Albert221, line 39: What exceptions could we catch here? Should we return `null` in this case anyway?

packages/login_client_flutter/lib/src/flutter_web_secure_credentials_storage.dart
```diff
@@ -0,0 +1,49 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// ignore: avoid_web_libraries_in_flutter
+import 'dart:html';
+
+import 'package:login_client/login_client.dart';
+
+/// A `flutter_secure_storage` implementation of the [CredentialsStorage].
+class FlutterSecureCredentialsStorage implements CredentialsStorage {
```
Albert221, line 21: Actually, LocalStorage is not secure storage. I understand you named this exactly as the iOS/Android class, but I believe this should not be solved that way.

Albert221 (changes requested):
The web storage implementation is okay, although I don't think we should hide those implementations in a class that suggest the use of `flutter_secure_storage`. 

IMO what should be done is leave the stub for the `FlutterSecureCredentialsStorage` class, but remove the `dart.library.js` case here. We should also create a separate LocalStorage class (like `WebLocalCredentialsStorage` or something similar) that would also have a stub.

Then we could create an umbrella storage that uses one of those two, depending on the platform. And that would utilize such an import for `dart.library.io`, `dart.library.js`, and the stub. This could be named e.g. `UmbrellaCredentialsStorage`.

packages/login_client_flutter/lib/src/flutter_web_secure_credentials_storage.dart
FlowerAlex: If I used another name for this class, I should do this for mobile also. They should have equal names for conditional import

packages/login_client_flutter/lib/src/flutter_web_secure_credentials_storage.dart
FlowerAlex: I think in the web app you have always access to local storage. If credentials are null, just return null. But I added this cause I wasn't sure about errors using getter 'localStorage'

Albert221 (approved):
LGTM! :+1:

Albert221:
Turns out the [`flutter_secure_storage` package](https://pub.dev/packages/flutter_secure_storage) introduced Web (and other platforms) support literally a few days ago, currently released as a prerelease. It uses [`SubtleCrypto` API](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto).

[Data on support for the mdn-api__SubtleCrypto feature across the major browsers from caniuse.com](https://caniuse.com/mdn-api_subtlecrypto)

Albert221:
flutter_secure_storage landed Web support. I propose we close this PR. @mateuszwojtczak @FlowerAlex

mateuszwojtczak:
sounds good! let's go

PR #38, opened 2021-11-19
Increase flutter_secure_storage version range

### LINKED ISSUE #31: Add web & umbrella credentials storage
Create the `CredentialsStorage` implementation for:

* the Web platform that doesn't have a dependency on the `dart:io` or iOS/Android platforms and uses [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) as its backend
* both mobile (iOS/Android) and Web platforms that's just an umbrella for the credentials storage above and the `FlutterSecureCredentialsStorage`. Probably involving [Dart's conditional importing](https://dart.dev/guides/libraries/create-library-packages#conditionally-importing-and-exporting-library-files).

/cc @mchudy @FlowerAlex

shilangyu:
Actually closes #31 

As far as I can tell the API has not changed, tested 4.2.0 and 5.0.0 and both seem to work.

### DIFF
--- a/packages/login_client_flutter/pubspec.yaml
+++ b/packages/login_client_flutter/pubspec.yaml
@@ -1,5 +1,5 @@
 name: login_client_flutter
-version: 2.0.0
+version: 2.0.1
 homepage: https://github.com/leancodepl/flutter_corelibrary/tree/master/packages/login_client_flutter
 repository: https://github.com/leancodepl/flutter_corelibrary
 description: >-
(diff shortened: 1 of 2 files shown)


packages/login_client_flutter/pubspec.yaml
```diff
@@ -13,7 +13,7 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  flutter_secure_storage: ^4.2.0
+  flutter_secure_storage: ">=4.2.0 <6.0.0"
```
Albert221, line 16: Please use single quotes

shilangyu: Right, sorry, autoformatter

Albert221 (approved):

PR #40, opened 2021-11-22
Fix lint rules

bartekpacia:
Are we:
-  simply not including unused lint rules 
or 
- commenting out unused lint rules with some rationale/comment

@RobertOdrowazLNCD Should we move `flutter_lints` from `dev_dependencies` to `dependencies`?

### DIFF
--- a/packages/cqrs/pubspec.yaml
+++ b/packages/cqrs/pubspec.yaml
@@ -7,7 +7,7 @@ description: >-
   queries and commands.
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
+  sdk: ">=2.12.0 <3.0.0"
 
 dependencies:
   http: ^0.13.3
--- a/packages/cqrs/pubspec.yaml
+++ b/packages/cqrs/pubspec.yaml
@@ -7,7 +7,7 @@ description: >-
   queries and commands.
 
 environment:
-  sdk: ">=2.12.0 <3.0.0"
+  sdk: '>=2.12.0 <3.0.0'
 
 dependencies:
   http: ^0.13.3
--- a/packages/leancode_lint/lib/analysis_options.yaml
+++ b/packages/leancode_lint/lib/analysis_options.yaml
@@ -186,7 +186,8 @@ linter:
     # Included in lints recommended
     # - library_prefixes
 
-    - library_private_types_in_public_api
+    # Is angry at private StatefulWidget's state
+    # - library_private_types_in_public_api
 
     # - lines_longer_than_80_chars
 
--- a/packages/leancode_lint/lib/analysis_options.yaml
+++ b/packages/leancode_lint/lib/analysis_options.yaml
@@ -186,9 +186,6 @@ linter:
     # Included in lints recommended
     # - library_prefixes
 
-    # Is angry at private StatefulWidget's state
-    # - library_private_types_in_public_api
-
     # - lines_longer_than_80_chars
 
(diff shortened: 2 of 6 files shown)


packages/cqrs/pubspec.yaml
```diff
@@ -7,7 +7,7 @@ description: >-
   queries and commands.
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
+  sdk: ">=2.12.0 <3.0.0"
```
shilangyu, line 10: Afaik we are sticking to single quotes

bartekpacia: VSCode changed it for me, sorry.

Looks like [Flutter prefers double quotes](https://docs.flutter.dev/development/packages-and-plugins/developing-packages#plugin-platforms) and [Dart prefers single quotes](https://dart.dev/tools/pub/pubspec#sdk-constraints) (or it might be just a coincidence and there's no consensus over this, so let's not bikeshed).

packages/leancode_lint/lib/analysis_options.yaml
```diff
@@ -280,7 +281,9 @@ linter:
 
     - prefer_final_in_for_each
     - prefer_final_locals
-    - prefer_final_parameters
+
+    # Not useful and annoying
+    # - prefer_final_parameters
```
Albert221, line 286: I'd simply remove it, not comment

packages/login_client/pubspec.yaml
```diff
@@ -8,7 +8,7 @@ description: >-
   refreshes tokens.
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
+  sdk: ">=2.12.0 <3.0.0"
```
Albert221: Why are you changing those quotes?

Albert221 (changes requested):

packages/login_client/pubspec.yaml
bartekpacia: Explained above.

Albert221:
> Should we move flutter_lints from dev_dependencies to dependencies?

Why? If you're having `unawaited` in mind, we re-defined it in our package

bartekpacia:
@Albert221 I had this message on my mind:
Screenshot 2021-11-23 at 2 24 04 PM

packages/leancode_lint/lib/analysis_options.yaml
```diff
@@ -186,8 +186,6 @@ linter:
     # Included in lints recommended
     # - library_prefixes
 
-    - library_private_types_in_public_api
```
RobertOdrowazLNCD, line 189: Comment those rules out instead of removing them completely. I will be easier to track new lints being added to dart in the future

RobertOdrowazLNCD (changes requested):

packages/leancode_lint/lib/analysis_options.yaml
Albert221: Eh, now I realized that actually all disabled lints are here commented out. Yup, let's leave it then, but without the comment on the annoyance :)

bartekpacia: The rabbis reached an agreement.

shilangyu (approved):

RobertOdrowazLNCD (approved):

Albert221 (approved):

PR #44, opened 2021-11-24
Use `leancode_lint`

### LINKED ISSUE #43: Use `leancode_lint` in packages
Currently all our packages are using [lint](https://pub.dev/packages/lint). 
Do we want to replace it with [leancode_lint](https://pub.dev/packages/leancode_lint)?

bartekpacia:
fix #43


packages/cqrs/analysis_options.yaml
```diff
@@ -1,5 +1,6 @@
-include: package:lint/analysis_options_package.yaml
+include: package:leancode_lint/analysis_options.yaml
```
shilangyu, line 1: You might want to use the `_package` variant

shilangyu:
I don't think `leancode_lint` should be depending on flutter

packages/cqrs/pubspec.yaml
```diff
@@ -7,10 +7,11 @@ description: >-
   queries and commands.
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
+  sdk: ">=2.12.0 <3.0.0"
```
RobertOdrowazLNCD, line 10: In PR #40 the quotes were changed the other way around in `leancode_lint` pubspec. We have to decide on some convention for it

packages/login_client/pubspec.yaml
```diff
@@ -13,12 +13,12 @@ environment:
 dependencies:
   http: ^0.13.0
   http_parser: ^4.0.0
+  leancode_lint: ^1.0.1
```
RobertOdrowazLNCD, line 16: I think it can be a dev dependency

packages/cqrs/pubspec.yaml
bartekpacia: ugh, autoformatter.

sorry

But TBH I don't know why are we forcing single quotes when the autoformatter itself uses double quotes. [some tiny research here](https://github.com/leancodepl/flutter_corelibrary/pull/40#discussion_r755114223)

packages/cqrs/analysis_options.yaml
bartekpacia: Why should I do this?
Also, after appending `_package` to this line, I'm getting warnings for lints that are not included in `leancode_lint`.
Screenshot 2021-11-28 at 2 46 59 PM

shilangyu: They are: https://github.com/leancodepl/flutter_corelibrary/blob/b00924d127ce80b1aefba703cd3f16d3919cfea7/packages/leancode_lint/lib/analysis_options_package.yaml#L6

the `_package` variant is supposed to be used for packages, cc @RobertOdrowazLNCD that was the point right?

packages/login_client/lib/src/login_client.dart
```diff
@@ -51,7 +51,7 @@ class LoginClient extends http.BaseClient {
   /// to correctly restore saved credentials from the `credentialsStorage`.
   ///
   /// See also:
-  /// - [InMemoryCredentialsStorage]
```
bartekpacia, line 54: Because of [this lint](https://dart-lang.github.io/linter/lints/comment_references.html)

Screenshot 2021-11-28 at 3 06 25 PM

packages/cqrs/analysis_options.yaml
bartekpacia: Yeah I found out after a while that there's also the `_package`. Makes perfect sense, thanks.

packages/login_client/lib/src/login_client.dart
bartekpacia: fix'd:)

bartekpacia:
I added docs to members where it was straightforward.

There are a few (7) more public memebrs left without documentation. I think that somebody more experienced should write docs for them. @Albert221 @shilangyu @RobertOdrowazLNCD

bartekpacia:
I added docs to all public members. Please criticise and help improve them:)

packages/login_client/lib/src/login_client.dart
```diff
@@ -206,11 +213,12 @@ class LoginClient extends http.BaseClient {
   /// Disposes the [LoginClient].
   Future<void> dispose() => _onCredentialsChanged.close();
 
-  // ignore: use_setters_to_change_properties
   @visibleForTesting
+  // ignore: use_setters_to_change_properties, public_member_api_docs
```
shilangyu, line 217: `use_setters_to_change_properties` does not have to be ignored here anymore, they probably fixed some false positives.

shilangyu (approved):

bartekpacia:
@shilangyu Looks like we have to drop support for Flutter 2.0.6 and 2.2.1.

shilangyu:
I think you can bump `leancode_lint` to `1.2.0` in all packages to remove the flutter dependency. Also, as you said, bumping CI's dart version to 2.14 sounds perfectly reasonable to me (and consequently dropping flutter `>=2.2.x`)

bartekpacia:
Oops, [this test](https://github.com/leancodepl/flutter_corelibrary/blob/master/packages/login_client/test/login_client_test.dart#L41-L57) is failing with Dart 2.14.

shilangyu:
It's not because of dart 2.14, its because of the `await` on `expectLater` (which has to wait on stream items which will be invoked later). `expect` without await could be used instead.

bartekpacia:
You're right, I forgot that a month ago I added `await` to all methods returning `Future`s (as I was instructed by `leancode_lint`), but actually not all calls to `expectLater` should be awaited. I wrapped these calls with `unawaited`.

Looks like it's fixed now :)

PR #54, opened 2022-04-28
add `package:leancode_hooks`

bartekpacia:
one package to hook 'em all


packages/leancode_hooks/lib/src/use_cooldown.dart
```diff
@@ -0,0 +1,35 @@
...
+/// Limits function calls to once per [duration].
+///
+/// When [CooledDown] function that this hook returns is called, a cooldown of
+/// [duration] is started. During it, any calls to [CooledDown] function are
+/// ignored.
+CooledDown useCooldown(
+  Duration duration, [
+  List<Object?> keys = const <Object>[],
+]) {
+  final debouncer = useMemoized(() => _Cooldowner(duration), keys);
```
kotertom, line 16: By the way, does the `_Cooldowner` ever need to be recreated? What do you think about dropping `keys` as parameter to `useCooldown` altogether and just pass `[duration]` as a single key to `useMemoized`?

packages/leancode_hooks/lib/src/use_synced_text_editing_controller.dart
```diff
@@ -0,0 +1,21 @@
+import 'package:flutter/widgets.dart';
+import 'package:flutter_hooks/flutter_hooks.dart';
+
+TextEditingController useSyncedTextEditingController(
+  void Function(TextEditingValue value) onChanged, {
+  String? initialText,
+}) {
+  final controller = useTextEditingController(text: initialText);
+
+  void listener() => onChanged(controller.value);
```
kotertom, line 10: How about moving the `listener` declaration inside the `useEffect`? It will be unnecessarily re-created every time the hook is called (not sure if this gets optimized during build)

packages/leancode_hooks/lib/src/use_cooldown.dart
```diff
@@ -0,0 +1,35 @@
...
+
+/// Signature for function that is returned from [useCooldown].
+typedef CooledDown = void Function(VoidCallback callback);
+
+/// Limits function calls to once per [duration].
+///
+/// When [CooledDown] function that this hook returns is called, a cooldown of
+/// [duration] is started. During it, any calls to [CooledDown] function are
+/// ignored.
+CooledDown useCooldown(
```
shilangyu, line 12: As we have discussed I think `useThrottle` would be more fitting

packages/leancode_hooks/lib/src/use_cooldown.dart
bartekpacia: It does not. I think that's a good idea, thank you :)

packages/leancode_hooks/lib/src/use_synced_text_editing_controller.dart
bartekpacia: thanks sir

packages/leancode_hooks/lib/src/use_google_map_controller.dart
```diff
@@ -0,0 +1,57 @@
...
+import 'package:flutter_hooks/flutter_hooks.dart';
+import 'package:google_maps_flutter/google_maps_flutter.dart';
+
+/// Returns a dark-mode aware [GoogleMapController].
+Completer<GoogleMapController> useGoogleMapController({
+  required String lightStylePath,
+  required String darkStylePath,
+}) {
+  final context = useContext();
+  final brightness = MediaQuery.of(context).platformBrightness;
```
shilangyu, line 14: Did you mean to use `Theme` here? Otherwise it won't react to app brightness changes and only to system wide ones

bartekpacia: Probably yes. In my app I tested it using system brightness only. Thanks.

bartekpacia: I am updating this `useGoogleMapController` hook to just get clever people to look at it. Then once it's considered "good" we can create a gist out of it and remove it from this package, so it won't be lost but also won't bloat the package.

bartekpacia:
@shilangyu FYI, I added another hook of yours – `useDeclarativeTextController`.

bartekpacia:
@kotertom I made `useDebounce` consistent with `useThrottle` (aka ex-`useCooldown`). It now returns a callback that can be used in many ways, one of which is to assign a value. Previosly assigning a value was the only use-case of debounce.

Please see [these changes](https://github.com/leancodepl/flutter_corelibrary/pull/54/commits/6ed909346287efa35a9f1c4f15cb998960db100e) and share your opinion

packages/leancode_hooks/lib/src/use_bloc.dart
```diff
@@ -0,0 +1,26 @@
...
+B useBloc<B extends Bloc<S, E>, S, E>(
+  B Function() create, [
+  List<Object> keys = const [],
+]) {
+  final bloc = useMemoized(create, keys);
+
+  useEffect(() => bloc.close, [bloc]);
+
+  return bloc;
+}
```
bartekpacia, line 14: Could we get off by removing generic parameters?

```dart
B useBloc<B extends Bloc<dynamic, dynamic>>(
  B Function() create, [
  List<Object> keys = const [],
]) {
  final bloc = useMemoized(create, keys);

  useEffect(() => bloc.close, [bloc]);

  return bloc;
}
```

packages/leancode_hooks/lib/src/use_stream_listener.dart
```diff
@@ -0,0 +1,31 @@
...
+        onError: onError,
+        onDone: onDone,
+        cancelOnError: cancelOnError,
+      );
+
+      return subscription.cancel;
+    },
+    keys,
+  );
+}
```
bartekpacia, line 31: I improved it in the way we talked about on weekly meeting on Friday. 

I thought that, given that `package:flutter_hooks` already has [useStream](https://pub.dev/documentation/flutter_hooks/latest/flutter_hooks/useStream.html) and [useStreamController](https://pub.dev/documentation/flutter_hooks/latest/flutter_hooks/useStreamController.html), we could contribute our `useStreamListener` to it.

[Looks like nobody has thought about it yet before :P](https://github.com/rrousselGit/flutter_hooks/issues?q=usestreamlistener)

packages/leancode_hooks/lib/src/use_decarative_text_editing_controller.dart
```diff
@@ -0,0 +1,22 @@
...
+          controller.text = text;
+        });
+      }
+      return null;
+    },
+    [text],
+  );
+
+  return controller;
+}
```
bartekpacia, line 22: @shilangyu Maybe we could merge this somehow with mine `useSyncedTextEditingController`?

I thought of using `useDeclarativeTextEditingController` inside `useSyncedTextEditingController`.

bartekpacia, line 22: @shilangyu Maybe we could merge this somehow with mine `useSyncedTextEditingController`?

I thought of using `useDeclarativeTextEditingController` inside `useSyncedTextEditingController`.e

packages/leancode_hooks/lib/src/use_focused.dart
```diff
@@ -0,0 +1,20 @@
...
+      }
+
+      focusNode.addListener(listener);
+      return () => focusNode.removeListener(listener);
+    },
+    [focusNode],
+  );
+
+  return focued.value;
+}
```
bartekpacia, line 20: @Albert221 this one is yours 😝

bartekpacia, line 20: @Albert221 this one is yours 😝

packages/leancode_hooks/lib/src/use_decarative_text_editing_controller.dart
bartekpacia, line 22: @shilangyu Maybe we could merge this somehow with mine `useSyncedTextEditingController`?

I thought of using `useDeclarativeTextEditingController` inside `useSyncedTextEditingController`.e

packages/leancode_hooks/lib/src/use_bloc.dart
bartekpacia: this is great, thanks

packages/leancode_hooks/lib/src/use_focused.dart
bartekpacia: It seems they do the same. 

In this case I'd ditch this hook.

packages/leancode_hooks/lib/src/use_decarative_text_editing_controller.dart
bartekpacia: Would you like to refactor `useDeclarativeTextEditingController` to also handle these edge cases or should we ditch it?

 To me, it looks like a useful hook.

packages/leancode_hooks/lib/src/use_focused.dart
shilangyu: IMO there is no need for a hook for this, `useListenable` covers this usecase (and it's what I have been using)

packages/leancode_hooks/lib/src/use_decarative_text_editing_controller.dart
shilangyu: TBH I dont understand what `useSyncedTextEditingController` does, looks like it only adds a callback. I though it would do what `useDeclarativeTextEditingController` aimed to do

bartekpacia: It's basically a common hook wrapping `useTextEditingController` with adding a listener:

I found myself writing this quite often
```dart
  final controller = useTextEditingController(text: initialText);

  void listener() => onChanged(controller.value);

  useEffect(
    () {
      controller.addListener(listener);
      return () => controller.removeListener(listener);
    },
    [controller, onChanged],
  );

  return controller;
```

so I extracted this to a hook.

I'd say that for `v0.1.0` we can keep this.

bartekpacia:
I'd like to keep `useGoogleMapController` hook in `v0.1.0` and remove it just afterwards (in `v0.1.1`), so it stays in git forever.

packages/leancode_hooks/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2021 LeanCode Sp. z o.o.
```
Albert221, line 189: It's 2022.

packages/leancode_hooks/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+[![leancode_hooks pub.dev badge][pub-badge]][pub-badge-link]
+
+[pub-badge]: https://img.shields.io/pub/v/leancode_hooks
+[pub-badge-link]: https://pub.dev/packages/leancode_hooks
+
+## Usage
+
+This package exports [`package:flutter_hooks`][flutter_hooks] so you won't have
+to depend on it.
```
Albert221, line 14: Could we have a list of the hooks we can use? Table with function name and a description would be very handy.

packages/leancode_hooks/lib/src/use_throttle.dart
```diff
@@ -0,0 +1,35 @@
...
+
+/// Returns a function that can be called only once per [duration].
+///
+/// When the returned function is called, a cooldown of [duration] is started.
+/// During its duration, any calls to the returned function are ignored.
+///
+/// See also:
+///
+/// * <https://rxmarbles.com/#throttle>, which this hook implements.
+ThrottledCallback useThrottle(Duration duration) {
```
Albert221, line 15: I like the name!

Albert221 (changes requested):

packages/leancode_hooks/README.md
bartekpacia: I think it's a good idea but it'd require more mainteneance work.

For `v0.0.1` I'd prefer a simple markdown list with links to files.

PR #62, opened 2022-05-17
Disable individual lints

bartekpacia:
This lets us disable individual lint rules.


shilangyu:
Maybe we just shouldn't base on `flutter_lint` instead of patching it?

bartekpacia:
@shilangyu I think it's worth considering.

I think that when we decided to base our lints on `flutter_lints`, we agreed with all the lints they had. But since they've included `use_build_context_synchronously_, this is not the case anymore.

PR #73, opened 2022-07-11
Remove outdated CQRS runtime nullability fix

shilangyu:


Albert221:
Let's also add a note to the readme, that this package assumes using a specific version of contracts generator or newer

shilangyu:
> Let's also add a note to the readme, that this package assumes using a specific version of contracts generator or newer
I would argue that this package should not be dependent on the contracts generator at all. Sure, we can make `package:cqrs` more convenient for the generator, but in the end of the day its the generator that should be adjusted to work with `package:cqrs`. So this version notice should be in the repo of contracts generator: but this is already enforced by adding `package:cqrs` dependency in the pubspec of the contracts generator.

Albert221 (approved):

PR #89, opened 2022-10-11
leancode_lint: bump version

bartekpacia:
Not sure what version to choose.

This can be potentially breaking if someone was depending on `unawaited()` from `package:leancode_lint`. They'll have to use `unawaited()` from `dart:async` instead.

### DIFF
--- a/packages/leancode_lint/CHANGELOG.md
+++ b/packages/leancode_lint/CHANGELOG.md
@@ -1,7 +1,8 @@
-## Unreleased
+## 2.0.0
 
+- **Breaking:** Remove the `unawaited()` function which is provided by Dart
+  since 2.14 (#88)
 - Disable the `discarded_futures` lint (#87)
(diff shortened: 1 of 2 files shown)


packages/leancode_lint/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-## Unreleased
+## 1.3.1
```
shilangyu, line 1: Defo 1.4.0 because of the unawaited removal

shilangyu (changes requested):

packages/leancode_lint/CHANGELOG.md
shilangyu: Actually, it should be 2.0.0 even

bartekpacia: thanks, I was inclined to do this as well but not sure

shilangyu (approved):

PR #92, opened 2022-12-19
Fix README build badges

Albert221:
* Fixed build badges in packages READMEs and in main README
* Instead of `beta` and `stable` channels, explicit latest stable version (3.3) is used in CI
* Make Codecov informational, so it won't fail the CI


bartekpacia:
I liked having `beta`, it helps us prepare for changes in new Flutter versions.

If not in the as a PR hook, we should make it running nightly.

shilangyu (approved):

PR #95, opened 2023-02-17
Use pub.dev automated publishing

Albert221:


bartekpacia (approved):
chad

shilangyu (commented):
Why are we bumping package versions?

Albert221:
So that a new package version is released checking if everything works correctly

PR #101, opened 2023-05-15
Remove dependence on flutter_lints

### LINKED ISSUE #98: Stop depending on `flutter_lints` in `leancode_lints`
The initial idea was to base our lints on `flutter_lints` to be compatible with community standards. But this has long stopped being the case. We now patch `flutter_lints` instead of extending it. For example we disable the `use_build_context_synchronously` rule since we deemed it too restrictive.

I suggest we drop `flutter_lints` completely and just maintain our own set of rules. One less dependency also means we don't have to be wary of upgrading `flutter_lints` in fear that it might introduce/disable some unwanted/wanted rules.

shilangyu:
closes #98


bartekpacia (approved):
Idea: I'd like to have an easy way to diff the rules enabled by `flutter_lints` and by `leancode_lint`.

PR #105, opened 2023-06-01
Bump package:http to v1 in packages

shilangyu:
Since we are requiring dart 3 here, maybe it is worth revisiting the API to modernize for records/class modifiers?

### DIFF
--- a/.github/workflows/cqrs-publish.yml
+++ b/.github/workflows/cqrs-publish.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Setup Dart
         uses: dart-lang/setup-dart@v1
         with:
-          sdk: 2.18.6
+          sdk: 3.0.1
 
       - name: Publish
         run: dart pub publish -f
--- a/packages/cqrs/CHANGELOG.md
+++ b/packages/cqrs/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 9.0.0
+
+- Bumped `http` dependency to `1.0.0`.
+
 # 8.0.0+1
 
 - Fix build badge in README.
--- a/packages/cqrs/CHANGELOG.md
+++ b/packages/cqrs/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 9.0.0
 
 - Bumped `http` dependency to `1.0.0`.
+- **Breaking:** Bump minimum Dart version to 3.0.
 
 # 8.0.0+1
 
--- a/packages/cqrs/CHANGELOG.md
+++ b/packages/cqrs/CHANGELOG.md
@@ -1,7 +1,7 @@
 # 9.0.0
 
-- Bumped `http` dependency to `1.0.0`.
-- **Breaking:** Bump minimum Dart version to 3.0.
+- Bumped `http` dependency to `1.0.0`. (#105)
+- **Breaking:** Bump minimum Dart version to 3.0. (#105)
 
 # 8.0.0+1
 
--- a/packages/cqrs/pubspec.yaml
+++ b/packages/cqrs/pubspec.yaml
@@ -7,14 +7,14 @@ description: >-
   queries and commands.
 
 environment:
-  sdk: ">=2.17.0 <3.0.0"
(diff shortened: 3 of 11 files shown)


.github/workflows/cqrs-publish.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       - name: Setup Dart
         uses: dart-lang/setup-dart@v1
         with:
-          sdk: 2.18.6
+          sdk: 3.0.1
```
bartekpacia, line 27: Here you don't use quotes, but in other places you do.
Is there any difference?

packages/cqrs/CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 9.0.0
+
+- Bumped `http` dependency to `1.0.0`.
+- **Breaking:** Bump minimum Dart version to 3.0.
```
bartekpacia, line 4: I like linking PRs in changelogs, like this:

```
- Bumped `http` dependency to `1.0.0` (#105)
- **Breaking:** Bump minimum Dart version to 3.0 (#105)
```

But that's just my quirk, I guess.

packages/cqrs/pubspec.yaml
```diff
@@ -1,20 +1,20 @@
...
+version: 9.0.0
 homepage: https://github.com/leancodepl/flutter_corelibrary/tree/master/packages/cqrs
 repository: https://github.com/leancodepl/flutter_corelibrary
 description: >-
   A library for convenient communication with CQRS-compatible backends, using
   queries and commands.
 
 environment:
-  sdk: ">=2.17.0 <3.0.0"
+  sdk: '>=3.0.1 <4.0.0'
```
bartekpacia, line 10: why 3.0.1 instead of 3.0.0?

packages/login_client/pubspec.yaml
```diff
@@ -8,17 +8,17 @@ description: >-
   refreshes tokens.
 
 environment:
-  sdk: '>=2.12.0 <3.0.0'
+  sdk: '>=3.0.1 <4.0.0'
```
bartekpacia, line 11: ditto

bartekpacia (commented):
some comments

packages/cqrs/pubspec.yaml
shilangyu: Formatting difference

packages/login_client/pubspec.yaml
shilangyu: As above

.github/workflows/cqrs-publish.yml
shilangyu: I am no yaml expert, but I think both produce strings? I just replaced whatever was there already. We can change every occurrence to be consistent

packages/cqrs/CHANGELOG.md
shilangyu: It is definitely a good idea, easier to track down the change

.github/workflows/cqrs-publish.yml
bartekpacia: I prefer to not use quotes in YAML wherever possible, but when it comes to semver versions such as this, I like to put single quotes around it. I recall there are some wild problems with sth like `0.2.3` being treated as octal (I honestly don't remember).

shilangyu: Would you like me to change all occurrences of version strings to be quoted?

Albert221:
`coverage` 1.3.0 change:
> Add a --package flag, which takes the package's root directory, instead of the .package file. Deprecate the --packages flag.

.github/workflows/cqrs-publish.yml
shilangyu: friendly ping @bartekpacia

bartekpacia: Sorry for the lag.

I'd prefer to not use quotes at all, and only add them if something breaks.

bartekpacia (approved):

PR #107, opened 2023-06-02
Modernize package:cqrs for dart 3

shilangyu:


packages/cqrs/lib/src/transport_types.dart
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 /// Base class for contracts that can be serialized and sent to the backend.
```
Albert221, line 15:
```suggestion
/// An interface for contracts that can be serialized and sent to the backend.
```

packages/cqrs/lib/src/transport_types.dart
```diff
@@ -13,7 +13,7 @@
...
-abstract class CqrsMethod {
+abstract interface class CqrsMethod {
   /// Returns a JSON-encoded representation of the data this class carries.
   Map<String, dynamic> toJson();
```
Albert221, line 19: Below, in the Dartdoc for `getFullName()`, we use "contractable" word, which was the name of this interface before. Now it's `CqrsMethod`, so maybe we can update it

packages/cqrs/lib/src/transport_types.dart
```diff
@@ -23,16 +23,16 @@ abstract class CqrsMethod {
 }
 
 /// Query describing a criteria for a query and the results it returns.
-abstract class Query<T> implements CqrsMethod {
+abstract interface class Query<T> implements CqrsMethod {
```
Albert221, line 26: can this interface extend that interface? I'm not sure whether it makes any real difference, but interfaces usually extend other interfaces 🤔

packages/cqrs/test/cqrs_test.dart
```diff
@@ -234,16 +232,17 @@ void main() {
 }
 
 void mockClientPost(MockClient client, Response response) {
+  registerFallbackValue(Uri());
```
Albert221, line 235: Could we register it just at the top of `main` in `setUpAll`?

packages/cqrs/lib/src/transport_types.dart
shilangyu: Changed, I don't think it will make any difference. But you are right, it makes more sense to extend.

shilangyu:
> I think we may also make an interface for the `Cqrs` API and make the current `Cqrs` class final and implement that interface. The interface could be abstracted out to the actual transport type (if we'll ever want to use WebSockets for transporting CqrsMethods?) and make it possible to mock the cqrs client. It would be breaking though.

While a nice idea I am not sure it will have any uses (at least not yet). Maybe we should wait for a need to popup and then make a new breaking change?

Albert221 (approved):

PR #111, opened 2023-06-16
chore: update melos to 3.0

### LINKED ISSUE #110: Migrate to melos 3
There's a newer melos version that isn't backwards compatible and requires changes to melos.yaml + adding workspace-level pubspec. I'm not sure how many people still use melos 2 but it'd be nice to not have to swap versions for those who already have v3 installed

https://melos.invertase.dev/guides/migrations#200-to-300

kotertom:
Resolves #110. All contributors should upgrade their melos to version 3.1. For codebase migration steps refer to https://melos.invertase.dev/guides/migrations#200-to-300


.gitignore
```diff
@@ -0,0 +1,71 @@
...
+
+# See https://www.dartlang.org/guides/libraries/private-files
+
+# Files and directories created by pub
+.dart_tool/
+.packages
+build/
+# If you're building an application, you may want to check-in your pubspec.lock
+pubspec.lock
+
```
shilangyu, line 12: Possibly everything below this line should be removed and individual package should decide what they want to ignore. But I don't know what is the generally accepted approach in such situations

kotertom: Sure, I agree. Let's keep this gitignore as light as possible. I'll remove everything that isn't related to this monorepo root/melos itself

PR #112, opened 2023-06-19
Add the enhanced_gradients package

kotertom:
I'd like to add a new package -- enhanced_gradients. As the name suggests, it allows us to automatically "enhance" (subjectively) gradients we create. Please refer to the readme for a short overview of how it's supposed to be used. If/once the package itself is ok, could you help me set up GH actions for it and publish it?


packages/enhanced_gradients/README.md
```diff
@@ -0,0 +1,46 @@
+# enhanced_gradients
+
+This package adds an easy way to enhance your Flutter gradients by interpolating colors
```
Albert221, line 3: It's not only for gradients but also for any color transitions, most commonly components' state changes

bartekpacia:
I'm not sure if this should be in `flutter_corelibrary`. I don't think gradients are "core". IMHO, e.g. `bloc_presentation` is more "core", but it's in a separate repo.

> could you help me set up GH actions for it and publish it?

Of course!

packages/enhanced_gradients/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+EnhancedLinearGradient(
+  colors: const [Color(0xFF000000), Color(0xFFFFFFFF)],
+  // ...
+)
+
+EnhancedRadialGradient(
+  colors: const [Color(0xFF000000), Color(0xFFFFFFFF)],
+  // ...
+).enhanced()
```
Albert221, line 40: Those extension methods shouldn't be here I suppose

packages/enhanced_gradients/README.md
kotertom: Well you're right! I've completely forgotten about this use case (and the HctColorTween from which the entire project started)

packages/enhanced_gradients/README.md
kotertom: Whoopsie 😅

kotertom:
@bartekpacia 

> I'm not sure if this should be in `flutter_corelibrary`. I don't think gradients are "core". IMHO, e.g. `bloc_presentation` is more "core", but it's in a separate repo. No strong opinions though.

I'm having these doubts as well, hence my vague slack thread on friday. Ultimately I thought it would be much easier to manage all these packages in a single monorepo but perhaps having at least a separate UI-focused goodies repository would be more beneficial.

mkucharski17:
> 
Imo approach of this repo is closer to "flutter_common_library" than "flutter_core_library", so I think adding this package is ok. I understand "core" as smth that is crucial and required. I would suggest changing the name of the repo to "flutter_common_library". On the other hand, if we want to stay with this name I think having "flutter_common_library" is a good idea. All packages like bloc_presentation/leancode_hooks etc. suit there. 

Do we have written down somewhere the approach of this repo? If not, I think it would be great to have it. Otherwise could have a discussion on all future packages.

bartekpacia:
I'm renaming this repo and keeping all our Dart/Flutter packages here. Another idea for the name `flutter_packages`.


Some examples from the community:
- [AboutYou has a monorepo repo called `dart_packages`](https://github.com/aboutyou/dart_packages)
- [VGV has a single repo per package](https://github.com/verygoodopensource)

kotertom:
Ok, so here goes a little summary from me as I see it. There are 3 solutions: 

1. a single monorepo for all our public packages
2. a monorepo per thematically related set of packages (like auth, cqrs, ui)
3. a repo per package

In all three cases we can use melos, so automated versioning and publishing that it provides applies to every approach.

## Single monorepo for all our public packages:

### Pros:

- automatic pub/path linking between dependent packages via melos (i.e. [what `melos bootstrap` does](https://melos.invertase.dev/commands/bootstrap#why-is-bootstrapping-required)) 
- code colocation for all our packages
- if any package at any point starts depending on another, it will be easy to setup local path overrides

### Cons:

- if the monorepo grows large (semi-plausible) we'll be downloading and sifting through tons of unrelated code
- similarily, if the monorepo grows large the dart analysis server might be prone to memory leaks; vide one of our commercial projects, where everyone complains about not being able to work efficiently without workarounds like only opening specific packages in their IDE
- work in progress packages are instantly public when pushed to remote (not sure if this is ever a problem)
- I'm worried about handling completely unrelated issues at some potential point in future

## Thematic monorepo

Theme examples: auth, network layer, UI, CQRS, package with a dart "core" + flutter "adapter/wrapper" layer

### Pros:

- same as with single monorepo, automatic local path package linking
- code colocation for related packages
- better visibility control (public/private)

### Cons:

- if a package at some point needs to start depending on a package from another monorepo, it will have to use the published version as a dependency (or will require manual setup), making working on both packages cumbersome
- how do you establish theme boundaries?

## Repo per package

### Pros:

- most granular visibility control (public/private) -- I see this advantage as pretty useless, since if you e.g. have a dart "core" package and a flutter wrapper, you probably want both to be either public or private
- lightweight

### Cons:

- hard to work on multiple interdependent packages at the same time
- [subjective] clutters organization repository list

Honestly, the repo-per-package approach seems to me the least appealing, since there are hardly any advantages and it seems cumbersome to work with without manual setup. The huge monorepo in turn seems to be pretty easy to handle but I'm seriously worried about it growing too large and crashing IDEs and the issue tab in github becoming a mess. The thematic monorepo seems really fine in theory, but in practice I find it hard to determine when a package should warrant a new repository, and when it should be part of another (and which one).

bartekpacia:
I like the first option most (single monorepo for all pkgs).

packages/enhanced_gradients/pubspec.yaml
```diff
@@ -0,0 +1,23 @@
...
+  collection: ^1.17.1
+  flutter:
+    sdk: flutter
+  material_color_utilities: ^0.2.0
+
+dev_dependencies:
+  flutter_test:
+    sdk: flutter
+  glados: ^1.1.6
+  leancode_lint: ">=1.0.2 <1.1.0"
```
Albert221, line 22: We can use latest now

packages/enhanced_gradients/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 0.0.1
+
+* TODO: Describe initial release.
```
Albert221, line 3: todo here

packages/enhanced_gradients/README.md
```diff
@@ -0,0 +1,49 @@
+# enhanced_gradients
+
+This package adds an easy way to enhance your Flutter gradients and color transitions by interpolating colors
+in the HCT color system provided by the [material_color_utilities package](https://pub.dev/packages/material_color_utilities).
+
+## Usage
+
+The package exposes two ways to modify the builtin `LinearGradient`, `RadialGradient` and `SweepGradient` gradients:
```
bartekpacia, line 8:
```suggestion
The package exposes two ways to modify the built-in `LinearGradient`, `RadialGradient` and `SweepGradient` gradients:
```

packages/enhanced_gradients/CHANGELOG.md
kotertom: I'm a bit confused as to how to properly write an initial changelog :/ Also, do we use or want to use `melos version`?

packages/enhanced_gradients/README.md
kotertom: :thonk: is the non-separated version really incorrect? I'm ok with either, so sure let's change it to built-in

.github/workflows/enhanced_gradients-publish.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Setup Dart
+        uses: dart-lang/setup-dart@v1
+        with:
+          sdk: 3.0.0
+
+      - name: Set up Flutter
+        uses: subosito/flutter-action@v2
+        with:
+          flutter-version: 3.10.0
```
bartekpacia, line 33:
```suggestion
flutter-version: 3.10.x
```

a bit sad that [Dart action doesn't support this](https://github.com/dart-lang/setup-dart/issues/56)

.github/workflows/enhanced_gradients-test.yml
```diff
@@ -0,0 +1,51 @@
...
+
+    defaults:
+      run:
+        working-directory: packages/enhanced_gradients
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Flutter
```
bartekpacia, line 33:
```suggestion
      - name: Set up Flutter
```

bartekpacia (approved):

packages/enhanced_gradients/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 1.0.0-rc1
```
Albert221, line 1: We'll release 1.0.0 after merge I suppose

packages/enhanced_gradients/pubspec.yaml
```diff
@@ -0,0 +1,23 @@
+name: enhanced_gradients
+version: 1.0.0-rc1
```
Albert221, line 2: Same here

packages/enhanced_gradients/pubspec.yaml
```diff
@@ -0,0 +1,23 @@
...
+  collection: ^1.17.1
+  flutter:
+    sdk: flutter
+  material_color_utilities: ^0.2.0
+
+dev_dependencies:
+  flutter_test:
+    sdk: flutter
+  glados: ^1.1.6
+  leancode_lint: ">=4.0.0 <4.1.0"
```
Albert221, line 22: `^4.0.0` is enough. Since `2.0.0` we started to bump major on SDK bumps

Albert221 (approved):

PR #113, opened 2023-06-28
Add example app

kotertom:
The project is generated with `flutter create --empty` (though I feel like some unnecessary comments still were generated), so the only file of note is lib/main.dart


kotertom:
🤔  somehow the test check attempts to run its scripts against the example project. Let me sort this out

bartekpacia:
Do we want to have the example app on all platforms? I sometimes like to do `flutter create --empty --platforms android,ios` (and maybe add `web` to that). More platforms can be very easily added later.

kotertom:
Umm, I haven't really given it much thought. I don't care much about supporting desktop right now and haven't checked if the app works there. Android and ios I agree are a must, and I'd like to keep web as well because it starts much faster, which is probably the most important in an example app.

bartekpacia (approved):
lgtm

bartekpacia (approved):

PR #115, opened 2023-07-04
[DebugPage] Log and display http requests

gogolon:
https://github.com/leancodepl/flutter_corelibrary/assets/31567861/6f50478a-44f0-4063-bafa-78d20908cd53


packages/debug_page/lib/src/ui/logs_inspector.dart
```diff
@@ -0,0 +1,43 @@
+import 'package:debug_page/debug_page.dart';
+import 'package:debug_page/src/request_log.dart';
+import 'package:debug_page/src/ui/details_screen/typography.dart';
+import 'package:debug_page/src/ui/log_tile.dart';
+import 'package:flutter/material.dart';
+
+class LogsInspector extends StatelessWidget {
```
RobertOdrowaz, line 7: We will definitely need to separate the logic from from there widget here. This page will not live all the time so it won't include requests which occurred when it wasn't visible

packages/debug_page/lib/src/ui/details_screen/typography.dart
```diff
@@ -0,0 +1,12 @@
+import 'package:flutter/material.dart';
+
+class DebugPageTypography {
+  static TextStyle _getTextStyle(double fontSize) => TextStyle(
```
RobertOdrowaz, line 4: I think having monospaced font for json etc. would be nice. Not in this PR just mentioning for the future

RobertOdrowaz (approved):

packages/debug_page/lib/src/ui/logs_inspector.dart
gogolon: true, I've already done it and it is coming in next PRs

PR #117, opened 2023-07-04
[DebugPage] Add entry points

gogolon:
#115  and #116  have to be merged first


packages/debug_page/example/ios/Podfile
```diff
@@ -0,0 +1,44 @@
+# Uncomment this line to define a global platform for your project
+# platform :ios, '11.0'
```
bartekpacia, line 2: let's make the lowest version explicit

packages/debug_page/lib/src/entry/debug_page_overlay.dart
```diff
@@ -0,0 +1,90 @@
...
+  @override
+  void dispose() {
+    super.dispose();
+    _loggerListener.dispose();
+    _shakeDetector?.stopListening();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    return Stack(
```
RobertOdrowaz, line 51: Add
```
fit: StackFit. passthrough
```
here

RobertOdrowaz (approved):

packages/debug_page/example/lib/util.dart
```diff
@@ -0,0 +1,28 @@
+import 'package:http/http.dart' as http;
+
+// Based on https://stackoverflow.com/a/51106300
+http.BaseRequest copyRequest(http.BaseRequest request) {
```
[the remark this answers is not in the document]
gogolon: we do, it's impossible to send a request stored in the same object twice

PR #126, opened 2023-07-13
Introduce LeanCodeFlutterSvgAdaptiveLoader

Xezolpl:


packages/leancode_flutter_svg_adaptive_loader/README.md
```diff
@@ -0,0 +1,74 @@
+# leancode_flutter_svg_adaptive_loader
+
+[![leancode_flutter_svg_adaptive_loader pub.dev badge][pub-badge]][pub-badge-link]
+[![][build-badge]][build-badge-link]
+
+This package adds an adaptive bytes loader for `.svg` assets. That allows you to compile your vector assets by `vector_graphics_compiler`
+into binary either inside of your repository or on CI/CD, without thinking in-code whether the asset is xml-based or a compiled binary.
```
Albert221, line 7: I'd add a sentence or two explaining that it allows you to provide a path to a file that can be either a normal SVG or this precompiled one using `vector_graphics_compiler`. The "without thinking" part is a little bit vague imo

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
```diff
@@ -0,0 +1,149 @@
...
+import 'package:flutter/widgets.dart';
+import 'package:flutter_svg/flutter_svg.dart';
+import 'package:vector_graphics/vector_graphics.dart';
+import 'package:vector_graphics_compiler/vector_graphics_compiler.dart' as vg;
+
+/// A [BytesLoader] able to distinguish xml-based svg from compiled binary form,
+/// by looking for '<svg' pattern at 1000 of first characters.
+/// Using logic from [AssetBytesLoader] for binaries and [SvgAssetLoader]
+/// for standard xml-based format.
+class LeancodeFlutterSvgAdaptiveLoader extends BytesLoader {
```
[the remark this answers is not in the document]
Xezolpl: Filename/folder name should be still `leancode` tho, not `lean_code` right? All of our packages use the first option

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
```diff
@@ -0,0 +1,149 @@
...
+    return _AssetByteLoaderCacheKey(
+      assetName,
+      packageName,
+      _resolveBundle(context),
+    );
+  }
+
+  @override
+  String toString() =>
+      'VectorGraphicAsset(${packageName != null ? '$packageName/' : ''}$assetName)';
```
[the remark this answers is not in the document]
Xezolpl: This name was used in flutter_svg code for both `AssetBytesLoader` and `SvgAssetLoader`, since our class is basically a copy of those with condition based on type I left it to be `VectorGraphicAsset`. Would you still recommend to rename it?

packages/leancode_flutter_svg_adaptive_loader/pubspec.yaml
```diff
@@ -0,0 +1,27 @@
...
+environment:
+  sdk: ">=3.0.0 <4.0.0"
+  flutter: ">=3.10.0"
+
+dependencies:
+  flutter:
+    sdk: flutter
+  flutter_svg: ^2.0.7
+  vector_graphics: ^1.1.7
+  vector_graphics_compiler: ^1.1.7
```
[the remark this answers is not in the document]
Xezolpl: If it would get removed, analyzer would show `depend_on_referenced_packages` in main src file

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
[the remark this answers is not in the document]
Albert221: Yes, leancode dir

packages/leancode_flutter_svg_adaptive_loader/pubspec.yaml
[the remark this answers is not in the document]
Albert221: Ah, right, so we do use it

packages/leancode_flutter_svg_adaptive_loader/pubspec.yaml
```diff
@@ -0,0 +1,27 @@
...
+dev_dependencies:
+  flutter_test:
+    sdk: flutter
+  leancode_lint: ^4.0.0
+  meta: ^1.9.1
+
+flutter:
+  assets:
+    - assets/foo.svg
+    - assets/foo_compiled.svg
```
shilangyu, line 27: Will that include these graphics for all users of this package?

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
```diff
@@ -0,0 +1,149 @@
...
+
+  /// This method intentionally avoids using `await` to avoid unnecessary event
+  /// loop turns. This is meant to to help tests in particular.
+  @override
+  Future<ByteData> loadBytes(BuildContext? context) {
+    return _prepareMessage(context).then((message) {
+      // Dart encodes strings in UTF-16 so one codepoint is 2 bytes
+      final first1000Chars = _decode(message, min(2000, message.lengthInBytes));
+
+      if (first1000Chars.contains('<svg')) {
```
shilangyu, line 61: Has this been studied to be reliable? It looks very hacky. I'll give it a shot to see if I can generate an example that breaks this check

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  Future<ByteData> _prepareMessage(BuildContext? context) {
+    return _resolveBundle(context).load(
+      packageName == null ? assetName : 'packages/$packageName/$assetName',
+    );
+  }
+
+  String _decode(ByteData? message, [int? lenght]) =>
+      utf8.decode(message!.buffer.asUint8List(0, lenght), allowMalformed: true);
```
shilangyu, line 51: Why are we asserting here?

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
```diff
@@ -0,0 +1,149 @@
...
+  /// This method intentionally avoids using `await` to avoid unnecessary event
+  /// loop turns. This is meant to to help tests in particular.
+  @override
+  Future<ByteData> loadBytes(BuildContext? context) {
+    return _prepareMessage(context).then((message) {
+      // Dart encodes strings in UTF-16 so one codepoint is 2 bytes
+      final first1000Chars = _decode(message, min(2000, message.lengthInBytes));
+
+      if (first1000Chars.contains('<svg')) {
+        return svg.cache.putIfAbsent(
```
shilangyu, line 62: Where is this `svg` object coming from? `putIfAbsent` sounds like a `Map` API, does it mean we are never controlling cache size and just pushing new caches?

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
```diff
@@ -0,0 +1,149 @@
...
+  ) {
+    return compute(
+      (message) {
+        return vg
+            .encodeSvg(
+              xml: _decode(message),
+              debugName: 'Svg loader',
+              enableClippingOptimizer: false,
+              enableMaskingOptimizer: false,
+              enableOverdrawOptimizer: false,
```
shilangyu, line 84: Is it slow? Does it generate artifacts? It produces large compile units? Why are we disabling optimizers?

packages/leancode_flutter_svg_adaptive_loader/pubspec.yaml
shilangyu, line 27: Are we not naming it `.svg.vec` on purpose?

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
shilangyu, line 15: I wonder if we necessarily want to name it with the `LeanCode...` prefix. AFAIK we don't do it anywhere, just name the package itself with the `leancode_` prefix.

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
```diff
@@ -0,0 +1,149 @@
...
+        other.packageName == packageName &&
+        other.assetBundle == assetBundle;
+  }
+}
+
+// Replaces the cache key for [AssetBytesLoader] to account for the fact that
+// different widgets may select a different asset bundle based on the return
+// value of `DefaultAssetBundle.of(context)`.
+@immutable
+class _AssetByteLoaderCacheKey {
```
shilangyu, line 123: Looks like it could be a simple record

packages/leancode_flutter_svg_adaptive_loader/README.md
```diff
@@ -0,0 +1,74 @@
...
+SvgPicture(LeancodeFlutterSvgAdaptiveLoader('assets/foo.svg')),
+```
+
+If you are using [flutter_gen package](https://pub.dev/packages/flutter_gen) you can copy the code below and use `.adaptiveSvg()` method being extension on `SvgGenImage`.
+
+```dart
+import 'package:leancode_flutter_svg_adaptive_loader/leancode_flutter_svg_adaptive_loader.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_svg/flutter_svg.dart';
+/// TODO: Add an import for `SvgGenImage` created by `flutter_gen`
```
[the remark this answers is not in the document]
shilangyu: resolved but not fixed

packages/leancode_flutter_svg_adaptive_loader/CHANGELOG.md
```diff
@@ -0,0 +1,3 @@
+## 1.0.0
+
+- Initial release with LeancodeFlutterSvgAdaptiveLoader
```
shilangyu, line 3: Wrap class name in backticks

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
```diff
@@ -0,0 +1,149 @@
...
+    return rootBundle;
+  }
+
+  Future<ByteData> _prepareMessage(BuildContext? context) {
+    return _resolveBundle(context).load(
+      packageName == null ? assetName : 'packages/$packageName/$assetName',
+    );
+  }
+
+  String _decode(ByteData? message, [int? lenght]) =>
```
shilangyu, line 50: typo
```suggestion
  String _decode(ByteData? message, [int? length]) =>
```

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
```diff
@@ -0,0 +1,149 @@
...
+  Future<ByteData> _loadSvg(
+    BuildContext? context,
+    ByteData? message,
+  ) {
+    return compute(
+      (message) {
+        return vg
+            .encodeSvg(
+              xml: _decode(message),
+              debugName: 'Svg loader',
```
shilangyu, line 81: This should identify the svg file. So maybe `'Svg loader: $filename'`

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
shilangyu: Follow up: check https://github.com/dnfield/vector_graphics/issues/210. Author is not keen on making the magic number public. But some other solutions were proposed instead (IMO all less than ideal ones). See if we can use any of them.

packages/leancode_flutter_svg_adaptive_loader/pubspec.yaml
Xezolpl: Yes, wanted to skip it to keep the same filename of assets for binary and xml-based forms.

packages/leancode_flutter_svg_adaptive_loader/README.md
Xezolpl: Updated, please verify if I explained it less vaguely :D

packages/leancode_flutter_svg_adaptive_loader/pubspec.yaml
Xezolpl: From my brief research:
1. Yes, including anything in package's `pubspec.yaml `will promote it also to users of the package.
2. There is no way of reading the asset file without having it defined in `pubspec.yaml`.
3. In Flutter framework tests they hack it by creating in test's `setUp` a new `pubspec.yaml`, writing by a buffer all the assets there and then testing what needs to be tested. https://github.com/flutter/flutter/blob/9cc4f94397f1f4b4d49ff38a02e5a0708dc9d68e/packages/flutter_tools/test/general.shard/asset_bundle_package_test.dart#L159

Xezolpl: From my brief research:
1. Yes, including anything in package's `pubspec.yaml `will promote it also to users of the package.
2. There is no way of reading the asset file without having it defined in `pubspec.yaml`.
3. In Flutter framework tests they hack it by creating in test's `setUp`

Xezolpl: I'd leave it attached since it is barely 2 KB

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
Xezolpl: That's how the author coded it in the `flutter_svg` package.

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
Xezolpl: That's how the author implemented it in the flutter_svg package. In my experience it can minify the compiled vector size.

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
Xezolpl: That's how the author implemented it in the `flutter_svg` package. 

That's how it looks in flutter_svg implementation

`/// Instance for [Svg]'s utility methods, which can produce a [DrawableRoot]
/// or [PictureInfo] from [String] or [Uint8List].
final Svg svg = Svg._();

/// A utility class for decoding SVG data to a [DrawableRoot] or a [PictureInfo].
///
/// These methods are used by [SvgPicture], but can also be directly used e.g.
/// to create a [DrawableRoot] you manipulate or render to your own [Canvas].
/// Access to this class is provided by the exported [svg] member.
class Svg {
  Svg._();

  /// A global override flag for [SvgPicture.cacheColorFilter].
  ///
  /// If this is null, the value in [SvgPicture.cacheColorFilter] is used. If it
  /// is not null, it will override that value.
  @Deprecated('This no longer does anything.')
  bool? cacheColorFilterOverride;

  /// The cache instance for decoded SVGs.
  final Cache cache = Cache();
}`

Xezolpl:
```

/// The cache for decoded SVGs.
class Cache {
  final Map<Object, Future<ByteData>> _pending = <Object, Future<ByteData>>{};
  final Map<Object, ByteData> _cache = <Object, ByteData>{};

  /// Maximum number of entries to store in the cache.
  ///
  /// Once this many entries have been cached, the least-recently-used entry is
  /// evicted when adding a new entry.
  int get maximumSize => _maximumSize;
  int _maximumSize = 100;

  /// Changes the maximum cache size.
  ///
  /// If the new size is smaller than the current number of elements, the
  /// extraneous elements are evicted immediately. Setting this to zero and then
  /// returning it to its original value will therefore immediately clear the
  /// cache.
  set maximumSize(int value) {
    assert(value != null); // ignore: unnecessary_null_comparison
    assert(value >= 0);
    if (value == maximumSize) {
      return;
    }
    _maximumSize = value;
    if (maximumSize == 0) {
      clear();
    } else {
      while (_cache.length > maximumSize) {
        _cache.remove(_cache.keys.first);
      }
    }
  }

  /// Evicts all entries from the cache.
  ///
  /// This is useful if, for instance, the root asset bundle has been updated
  /// and therefore new images must be obtained.
  void clear() {
    _cache.clear();
  }

  /// Evicts a single entry from the cache, returning true if successful.
  bool evict(Object key) {
    return _cache.remove(key) != null;
  }

  /// Evicts a single entry from the cache if the `oldData` and `newData` are
  /// incompatible.
  ///
  /// For example, if the theme has changed the current color and the picture
  /// uses current color, [evict] will be called.
  bool maybeEvict(Object key, SvgTheme oldData, SvgTheme newData) {
    return evict(key);
  }

  /// Returns the previously cached [PictureStream] for the given key, if available;
  /// if not, calls the given callback to obtain it first. In either case, the
  /// key is moved to the "most recently used" position.
  ///
  /// The arguments must not be null. The `loader` cannot return null.
  Future<ByteData> putIfAbsent(
    Object key,
    Future<ByteData> loader(),
  ) {
    assert(key != null); // ignore: unnecessary_null_comparison
    assert(loader != null); // ignore: unnecessary_null_comparison
    Future<ByteData>? pendingResult = _pending[key];
    if (pendingResult != null) {
      return pendingResult;
    }

    ByteData? result = _cache[key];
    if (result != null) {
      // Remove the provider from the list so that we can put it back in below
      // and thus move it to the end of the list.
      _cache.remove(key);
    } else {
      pendingResult = loader();
      _pending[key] = pendingResult;
      pendingResult.then((ByteData data) {
        _pending.remove(key);
        _add(key, data);
        result = data; // in case it was a synchronous future.
      });
    }
    if (result != null) {
      _add(key, result!);
      return SynchronousFuture<ByteData>(result!);
    }
    assert(_cache.length <= maximumSize);
    return pendingResult!;
  }

  void _add(Object key, ByteData result) {
    if (maximumSize > 0) {
      if (_cache.containsKey(key)) {
        _cache.remove(key); // update LRU.
      } else if (_cache.length == maximumSize && maximumSize > 0) {
        _cache.remove(_cache.keys.first);
      }
      assert(_cache.length < maximumSize);
      _cache[key] = result;
    }
    assert(_cache.length <= maximumSize);
  }

  /// The number of entries in the cache.
  int get count => _cache.length;
}

```

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
Xezolpl: Im not sure if our current implementation (looking for "<svg") requires any change? Do you think I should replace "<svg" with "<" alone?

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
shilangyu: If it can minify don't we want to do that?

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
```diff
@@ -0,0 +1,123 @@
...
+        );
+      } else {
+        return message;
+      }
+    });
+  }
+
+  Future<ByteData> _loadSvg(
+    BuildContext? context,
+    ByteData? message,
```
shilangyu, line 74: Context not used, and shouldnt message be non nullable? Not sure how we could possibly load it without the bytes. Then _decode could accept a non-nullable too

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
shilangyu: If we are going the `<` route it would be better if we lazily decoded utf8 to look for it, instead of arbitrarily getting 2000 chars. The 2000 chars can still be broken we we have a lot of leading whitespace. I can not find a way to lazily decode bytes to utf8 though. Maybe @Albert221 you have some ideas here?

packages/leancode_flutter_svg_adaptive_loader/pubspec.yaml
shilangyu: I think attaching any asset regardless of its size is unwanted. Mati's suggestion of having a separate package sounds good though?

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
Xezolpl: Sorry, didn't clarify, disabling those options can minify the size

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
Xezolpl: I believe it is a limitation that we can agree on, cuz it is very unlikely for some asset to have 1000 chars of white space / comments (only case I see is some kind of documentation for the vector)

Albert221: I'm thinking of a similar check: check first _n_ bytes for whether they're UTF8-encoded. If they are - treat it as SVG. If they're not - they're most probably the binary vector format

packages/leancode_flutter_svg_adaptive_loader/pubspec.yaml
Xezolpl: Tackled in this PR https://github.com/leancodepl/flutter_corelibrary/pull/141

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
Albert221: Yup, let's do this:

Read first _n_ bytes (let's make it 100), with utf8 decoder with `allowMalformed: true` and count `\uFFFD` characters (�; this is a character that the decoder places in place of invalid codepoints) in the code. If it's more than 1, we got invalid UTF-8 encoded string - we treat it as binary vector. If it's 1, we clipped the last codepoint when we read n bytes. It's fine. We treat it as SVG. If it's 0 - same story - we treat it as SVG.

packages/leancode_flutter_svg_adaptive_loader/lib/src/leancode_flutter_svg_adaptive_loader.dart
```diff
@@ -0,0 +1,120 @@
...
+import 'package:flutter/services.dart';
+import 'package:flutter/widgets.dart';
+import 'package:flutter_svg/flutter_svg.dart';
+import 'package:vector_graphics/vector_graphics.dart';
+import 'package:vector_graphics_compiler/vector_graphics_compiler.dart' as vg;
+
+/// A [BytesLoader] able to distinguish xml-based svg from compiled binary form,
+/// by looking for '<svg' pattern at 1000 of first characters.
+/// Using logic from [AssetBytesLoader] for binaries and [SvgAssetLoader]
+/// for standard xml-based format.
```
shilangyu, line 14: Remember to update the comment once the logic is changed

shilangyu (approved):

PR #127, opened 2023-07-13
Add example for LeancodeFlutterSvgAdaptiveLoader

Xezolpl:


pubspec.yaml
```diff
@@ -5,3 +5,5 @@ environment:
   sdk: ">=3.0.0 <4.0.0"
 dev_dependencies:
   melos: ^3.1.0
+dependencies:
+  flutter_svg: ^2.0.7
```
shilangyu, line 9: Did you mean to add this here?

shilangyu (approved):

pubspec.yaml
Xezolpl: Nope, thanks for catching that

PR #130, opened 2023-07-17
[Debug Page] Add warning regarding `Logger.root` to readme

gogolon:


packages/debug_page/README.md
```diff
@@ -61,4 +63,8 @@ class _MyAppState extends State<MyApp> {
 
 For a complete working sample, see [example](example).
 
-You can configure debug page's entry points by setting ```showEntryButton``` (defaults to false) and ```showOnShake``` (defaults to true) flags in the constructor of ```DebugPageController```.
\ No newline at end of file
+You can configure debug page's entry points by setting ```showEntryButton``` (defaults to false) and ```showOnShake``` (defaults to true) flags in the constructor of ```DebugPageController```.
+
+## Warning
+
+For gathering loggers logs, this package relies on listening to `Logger.root`. This means that changing `Logger.root.level` affects this package behavior.
```
shilangyu, line 70: Maybe also worth mentioning that it collects logs only from the current isolate

shilangyu (approved):

packages/debug_page/README.md
gogolon: isn't _main isolate_ more precise than _current_?

gogolon: right, I have already forgotten that it's the controller which starts listening 🙈

PR #131, opened 2023-07-18
[Force Update] V1

gogolon:


.gitignore
```diff
@@ -1,3 +1,6 @@
+# Miscellaneous
+.DS_Store
```
PiotrRogulski, line 2: Untrack all .DS_Stores

packages/force_update/backend/Class1.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace backend;
```
PiotrRogulski, line 1: Are those files supposed to be here? I think with the current configuration, the backend will be included as part of the Flutter package

gogolon: good question. I think it should be in repo, but definitely should not be in package. I'll think whether to remove it completely or just move the backend folder higher to make sure it's not included in the package

gogolon: moved backend folder up

packages/.gitignore
RobertOdrowaz, line 1: I don't think it's necessary since you already have it in root

packages/force_update/mobile/analysis_options.yaml
RobertOdrowaz, line 1: I think we should use leancode_lint here. Also add a new line at the end

packages/force_update/mobile/example/.gitignore
```diff
@@ -3,7 +3,7 @@
 *.log
 *.pyc
 *.swp
-.DS_Store
+**.DS_Store
```
RobertOdrowaz, line 6: I don't think you need to change this

packages/force_update/mobile/lib/src/app_version.dart
```diff
@@ -0,0 +1,42 @@
...
+  bool operator >(T other) => compareTo(other) > 0;
+}
+
+class AppVersion implements Comparable<AppVersion> {
+  const AppVersion({required this.version});
+
+  final String version;
+
+  @override
+  int compareTo(other) {
```
RobertOdrowaz, line 14: Using this package https://pub.dev/packages/version seems safer than writing this logic from scratch. I believe there are quite a lot of edge-cases here

RobertOdrowaz (approved):

packages/force_update/mobile/lib/src/app_version.dart
gogolon: replaced with the package

PR #139, opened 2023-08-03
Elevate most rules to warning level

RobertOdrowaz:
We don't want code with those infos on main branch so they should be warnings and cause CI checks to fail.

I'm not sure how we should approach semver in this case but seems like significant enough change to warrant major bump


shilangyu:
Shouldnt we instead setup CIs to just fail on infos? (thats what I am doing)

RobertOdrowaz:
There are things that shouldn't fail but should be listed ie. TODOs and deprecated (which I forgot to exclude) and warnings are pretty much unused and can be used for this

packages/leancode_lint/lib/analysis_options.yaml
```diff
@@ -3,6 +3,203 @@ analyzer:
...
+    deprecated_export_use: info
+    deprecated_member_use: info
+    division_optimization: warning
+    import_deferred_library_with_load_function: warning
+    import_of_legacy_library_into_null_safe: warning
+    unignorable_ignore: warning
+    unnecessary_cast: warning
+    unnecessary_import: warning
+    unreachable_switch_case: warning
+    unused_element: warning
```
shilangyu, line 202: hint?

RobertOdrowaz: Yep the ones from this file https://github.com/dart-lang/sdk/blob/main/pkg/analyzer/lib/src/dart/error/hint_codes.g.dart

shilangyu (approved):

PR #142, opened 2023-08-09
TapAnalyticsEvent need String instead of ValueKey

pdenert:
As in the title. We change the approach for passing analytics id key. Required to add analytics to leancode_flutter_template

### DIFF
--- a/packages/leancode_analytics/pubspec.yaml
+++ b/packages/leancode_analytics/pubspec.yaml
@@ -1,6 +1,6 @@
 name: leancode_analytics
 description: A new Flutter package project.
-version: 0.0.1
+version: 0.0.2
 homepage:
 
 environment:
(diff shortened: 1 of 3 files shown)


packages/leancode_analytics/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: leancode_analytics
 description: A new Flutter package project.
-version: 0.0.1
+version: 0.0.2
```
shilangyu, line 3: Its not published so i dont think you need to worry about it

shilangyu (approved):

packages/leancode_analytics/pubspec.yaml
pdenert: Yeah, but @KamilSztandur already using it in lncd project

PR #143, opened 2023-08-16
Enhanced gradients docs and example app

kotertom:
- enhanced_gradients@1.0.1


packages/enhanced_gradients/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.1
+
+ - **DOCS**(enhanced_gradients): expand readme and the example app.
```
PiotrRogulski, line 3: Since it's the changelog only for `enhanced_gradients`, I don't think specifying the name here is necessary

packages/enhanced_gradients/pubspec.yaml
```diff
@@ -13,7 +13,7 @@ dependencies:
   collection: ^1.17.1
   flutter:
     sdk: flutter
-  material_color_utilities: ^0.2.0
+  material_color_utilities: any
```
Albert221, line 16: Would our package really work with any version? We don't have test matrix for that. Let's add such

packages/enhanced_gradients/example/art/image.png
Albert221, line 1: Could we add more color pairs to the examples?

.github/workflows/enhanced_gradients-publish.yml
```diff
@@ -3,6 +3,8 @@ name: enhanced_gradients publish
 on:
   push:
     tags: ["enhanced_gradients-v*"]
+    branches:
+      - "master"
```
PiotrRogulski, line 7: Consider unifying the style of yaml arrays

CHANGELOG.md
shilangyu, line 1: What is the purpose of a top-level changelog?

.github/workflows/enhanced_gradients-publish.yml
kotertom: I've kept the tags style from the other configs. I think it does no harm as long as it reasonably fits in one line though

CHANGELOG.md
kotertom: Good question. I added it because I wanted to start using melos commands like `melos version` and it does generate this top-level changelog. I removed it now, as I don't see much of a point to keeping it, either

packages/enhanced_gradients/example/art/image.png
kotertom: Done!

packages/enhanced_gradients/pubspec.yaml
kotertom: Added (I think it works, we'll see)

packages/enhanced_gradients/CHANGELOG.md
kotertom: Done

.github/workflows/enhanced_gradients-test.yml
```diff
@@ -21,6 +24,7 @@ jobs:
       matrix:
         include:
           - version: "3.10.0"
```
Albert221, line 26: Could we add stable channel instead of 3.10.0?

Albert221 (approved):

shilangyu (approved):

shilangyu:
@kotertom is there a blocker here?

PR #148, opened 2023-08-23
Make "SafeCqrs" the default CQRS client

jtarkowski27:
### Goals
- Add and replace previous `Cqrs` with safe version that catches all possible exceptions and recognises list of errors returned from backend. 
- Add middleware mechanism that simplifies error handling globally across the app for certain errors.

### Questions
- Should we still keep previous version of `Cqrs` under some other name?
- Should I make error type not generic in all result classes i.e. in `CqrsQueryResult`, `CqrsCommandResult` and `CqrsOperationResult`?
- Should I shorten the names for most of the new classes i.e. `CqrsCommandResult` -> `CResult` etc?

### Usage porposal
```dart
// Define middleware
class ExampleMiddleware implements CqrsMiddleware {
  const ExampleMiddleware();

  @override
  Future<CqrsCommandResult<CqrsError>> handleCommandResult(
    CqrsCommandResult<CqrsError> result,
  ) {
    print('Is error: ${result.isError}');
    return Future.value(result);
  }

  @override
  Future<CqrsQueryResult<T, CqrsError>> handleQueryResult<T>(
    CqrsQueryResult<T, CqrsError> result,
  ) {
    print('Is error: ${result.isError}');
    return Future.value(result);
  }
}

// Provide and add middleware to safe Cqrs
Provider(
  create: (context) {
    final middleware = ExampleMiddleware();
    context.read<Cqrs>().addMiddleware(middleware);
    return middleware;
  },
  dispose: (context, middleware) {
    context.read<Cqrs>().removeMiddleware(middleware);
  },
  lazy: false,
  child: ...
),

// Handle validation errors for commands
final result = cqrs.run(ExampleCommand());
if (result case CommandSuccess()) {
  emit(SuccessState());
} else if (result case CcommandFailure(isInvalid: true, :final validationErrors)) {
  handleValidationErrors(validationErrors);
} 
```


shilangyu:
Extra package instead of something included in `package:cqrs`?

jtarkowski27:
@shilangyu I made it additional package since `CqrsWrapper` needs `equatable` and `logging` packages which are not used by the `cqrs` package. However we can consider moving it to the `cqrs` if it's okay to add those dependencies there.

shilangyu:
> @shilangyu I made it additional package since `CqrsWrapper` needs `equatable` and `logging` packages which are not used by the `cqrs`. However we can consider moving it to the `cqrs` if it's okay to add those dependencies there.

Personally I would prefer it to be in `cqrs`, but maybe I'm in the minority. I guess this isn't important for now, sorry for bringing it up.

packages/cqrs/lib/src/cqrs_result.dart
```diff
@@ -0,0 +1,202 @@
...
+
+  /// Whether this instance can be casted to [CqrsQuerySuccess].
+  bool get isSuccess => asSuccess != null;
+
+  /// Whether this instance can be casted to [CqrsQueryFailure].
+  bool get isFailure => asFailure != null;
+
+  /// Returns [data] of type [T] for [CqrsQuerySuccess] and `null`
+  /// for [CqrsQueryFailure].
+  T? get data => asSuccess?.data;
```
shilangyu, line 54: I am very much against this getter (and a little bit less against all other getters (asSuccess, asFailure, isSuccess, isFailure, as well as all other getters on `CqrsCommandResult`)).
- `T` could be a nullable type, using the `data` getter hides the distinction between "no data" and "data which is null". I don't believe this will be common knowledge and it will introduce bugs in codebases.
- we introduce type safety and exhaustiveness with sealed classes, but these getter throw it out of the window. Do we want people to be using `asSuccess` and ignoring the error case?

packages/cqrs/lib/src/cqrs_result.dart
```diff
@@ -0,0 +1,202 @@
...
+
+  /// Returns [error] of type [E] for [CqrsQueryFailure] and `null`
+  /// for [CqrsQuerySuccess].
+  E? get error => asFailure?.error;
```
shilangyu, line 58: Should error be generic? Isn't it always CqrsError?

packages/cqrs/lib/src/cqrs_result.dart
```diff
@@ -0,0 +1,202 @@
...
+
+  /// Whether this instance can be casted to [CqrsCommandFailure].
+  bool get isFailure => asFailure != null;
+
+  /// Returns an [error] of type [E] for [CqrsCommandFailure] and `null`
+  /// for [CqrsCommandSuccess].
+  E? get error => asFailure?.error;
+
+  /// Returns a list of [ValidationError] errors.
+  List<ValidationError> get validationErrors;
```
shilangyu, line 114: Any particular reason why we are not defining it only on the error variant?

jtarkowski27: I would say convenience. For command result it might be more handy to write:

```dart
if (result.isInvalid) {
  handleValidationErrors(result.validation)
}
```

than having to exhaust all posibble result types as so:

```dart
switch (result) {
  CqrsCommandFailure(
    isInvalid: true,
    :final validationErrors,
  ) =>
    handleValidationErrors(validationErrors),
  _ => null,
};
```

packages/cqrs/lib/src/cqrs_result.dart
jtarkowski27: That what I was thinking about too and I'm probably going to remove generic type for error.

packages/cqrs/lib/src/cqrs_result.dart
jtarkowski27: I do agree with the part regarding getters such as `data`, `error`, `asSuccess` etc. Note that previous installment of SafeCqrs had [the same `data` getter](https://github.com/leancodepl/leancode_flutter_template/blob/master/bricks/safe_cqrs/__brick__/%7B%7BprojectAbbr.snakeCase()%7D%7D_cqrs.dart#L83) which also ignored the mentioned conflict. 

However as for other getters i.e. `isSuccess`, `isFailure` and `isInvalid` I on the side of keeping them. We often want to check if a query has failed without precisely knowing what kind of error has happen. Same with checking if command was executed successfully.

packages/cqrs/lib/cqrs.dart
```diff
@@ -28,6 +28,13 @@
 /// // Fetching first page of flowers
 /// final flowers = await cqrs.get(AllFlowers(page: 1));
 ///
+/// // Handling query result
+/// if (result.isSuccesful) {
```
Albert221, line 32: successful is an adjective, failure is noun. please use one

packages/cqrs/lib/src/cqrs_result.dart
```diff
@@ -0,0 +1,202 @@
...
+/// [CqrsCommandResult] with error type [CqrsError].
+extension CqrsCommandResultValidationErrorExtension
+    on CqrsCommandResult<CqrsError> {
+  /// Whether is a [CqrsCommandFailure] with validation errors.
+  bool get isInvalid => error == CqrsError.validation;
+}
+
+/// Generic result for CQRS query result. Can be either [CqrsQuerySuccess]
+/// or [CqrsQueryFailure].
+sealed class CqrsQueryResult<T, E> extends Equatable {
```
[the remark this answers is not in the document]
jtarkowski27: I made it non generic.

packages/cqrs/lib/cqrs.dart
jtarkowski27: Fixed it.

packages/cqrs/lib/src/cqrs_result.dart
```diff
@@ -0,0 +1,160 @@
...
+  bool get isSuccess => switch (this) {
+        QSuccess<T>() => true,
+        _ => false,
+      };
+
+  /// Whether this instance is of final type [QFailure].
+  bool get isFailure => switch (this) {
+        QFailure<T>() => true,
+        _ => false,
+      };
```
shilangyu, line 36:
```suggestion
  bool get isSuccess => this is QSuccess<T>;

  /// Whether this instance is of final type [QFailure].
  bool get isFailure => this is QFailure<T>;
```

packages/cqrs/lib/src/cqrs_result.dart
```diff
@@ -0,0 +1,160 @@
...
+  bool get isSuccess => switch (this) {
+        CSuccess() => true,
+        _ => false,
+      };
+
+  /// Whether this instance is of final type [CFailure].
+  bool get isFailure => switch (this) {
+        CFailure() => true,
+        _ => false,
+      };
```
shilangyu, line 79: ditto

packages/cqrs/lib/src/cqrs_result.dart
```diff
@@ -0,0 +1,160 @@
...
+  bool get isSuccess => switch (this) {
+        OSuccess<T>() => true,
+        _ => false,
+      };
+
+  /// Whether this instance is of final type [OFailure].
+  bool get isFailure => switch (this) {
+        OFailure<T>() => true,
+        _ => false,
+      };
```
shilangyu, line 135: etc

packages/cqrs/lib/src/cqrs_error.dart
```diff
@@ -0,0 +1,31 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Error types that are recognized by Cqrs.
+enum CqrsError {
+  /// Represents a network/socket error
+  network,
+
+  /// Represents a HTTP 401 authetication error
```
shilangyu, line 20:
```suggestion
  /// Represents a HTTP 401 authentication error
```

packages/cqrs/lib/src/cqrs_error.dart
```diff
@@ -0,0 +1,31 @@
...
+  authentication,
+
+  /// Represents a HTTP 403 forbidden access error
+  forbiddenAccess,
+
+  /// Represents a HTTP 422 validation error
+  validation,
+
+  /// Represents a generic error which covers all remaining errors
+  unknown,
```
shilangyu, line 30:
```suggestion
  /// Represents a network/socket error.
  network,

  /// Represents a HTTP 401 authetication error.
  authentication,

  /// Represents a HTTP 403 forbidden access error.
  forbiddenAccess,

  /// Represents a HTTP 422 validation error.
  validation,

  /// Represents a generic error which covers all remaining errors.
  unknown,
```

packages/cqrs/lib/src/cqrs.dart
```diff
@@ -35,48 +59,62 @@ class Cqrs {
   /// The `timeout` defaults to 30 seconds. `headers` have lesser priority than
   /// those provided directly into [get] or [run] methods and will be overrided
   /// by those in case of some headers sharing the same key.
+  ///
+  /// Any result (be it successful of failure) of CQRS method will be logged
```
shilangyu, line 63:
```suggestion
  /// Any result (be it success or failure) of CQRS method will be logged
```

packages/cqrs/lib/src/cqrs.dart
```diff
@@ -35,48 +59,62 @@ class Cqrs {
...
+  /// given the `logger` is provided.
+  ///
+  /// In case when a global result handling is needed, one might provide
+  /// a `middlewares` list with a collection of [CqrsMiddleware] objects. Every
+  /// time a result is returned, [Cqrs.get] and [Cqrs.run] will execute for
+  /// each middleware on the list [CqrsMiddleware.handleQueryResult]
+  /// and [CqrsMiddleware.handleCommandResult] accordingly.
```
shilangyu, line 70: mention `perform`?

packages/cqrs/lib/src/cqrs.dart
```diff
@@ -166,4 +310,38 @@ class Cqrs {
...
+    final log = switch (result) {
+      _ResultType.success => _logger?.info,
+      _ResultType.validationError => _logger?.warning,
+      _ => _logger?.severe,
+    };
+
+    final methodTypePrefix = switch (method) {
+      Query() => 'Query',
+      Command() => 'Command',
+      _ => 'Operation',
```
shilangyu, line 330: Why is Operation type omitted?

packages/cqrs/lib/src/cqrs.dart
```diff
@@ -166,4 +310,38 @@ class Cqrs {
...
+    CqrsMethod method,
+    _ResultType result, [
+    Object? error,
+    StackTrace? stackTrace,
+    List<ValidationError> validationErrors = const [],
+  ]) {
+    final log = switch (result) {
+      _ResultType.success => _logger?.info,
+      _ResultType.validationError => _logger?.warning,
+      _ => _logger?.severe,
```
shilangyu, line 324: Personally I would write out all variants to be more explicit and get feedback when a new _ResultType is added

packages/cqrs/lib/src/cqrs.dart
```diff
@@ -166,4 +310,38 @@ class Cqrs {
...
   }
+
+  void _log(
+    CqrsMethod method,
+    _ResultType result, [
+    Object? error,
+    StackTrace? stackTrace,
+    List<ValidationError> validationErrors = const [],
+  ]) {
+    final log = switch (result) {
```
shilangyu, line 321: I'd early return if logger is null. All that work is for nothing if you have no logger.

packages/cqrs/lib/src/cqrs_result.dart
jtarkowski27: After all I removed the generic error type.

packages/cqrs/lib/src/cqrs_result.dart
jtarkowski27: In my recent changes I changed that only `CFailure` will have `validationErrors` as field.

packages/cqrs/lib/src/cqrs.dart
```diff
@@ -166,4 +311,47 @@ class Cqrs {
...
+  void _log(
+    CqrsMethod method,
+    _ResultType result, [
+    Object? error,
+    StackTrace? stackTrace,
+    List<ValidationError> validationErrors = const [],
+  ]) {
+    if (_logger == null) {
+      return;
+    }
```
shilangyu, line 324: Not exactly what I mean: since you are still being tormented with the nullable logger below. Pin to a local variable perhaps (`final logger = _logger`)?

packages/cqrs/lib/src/cqrs.dart
```diff
@@ -166,4 +311,47 @@ class Cqrs {
...
+      _ResultType.authenticationError ||
+      _ResultType.forbiddenAccessError ||
+      _ResultType.unknownError =>
+        _logger?.severe,
+    };
+
+    final methodTypePrefix = switch (method) {
+      Query() => 'Query',
+      Command() => 'Command',
+      Operation() || _ => 'Operation',
```
shilangyu, line 340: Perhaps we should mark `CqrsMethod` as sealed to avoid this catchall?

packages/cqrs/lib/src/cqrs.dart
shilangyu: or we can wait for dart 3.2 where private fields are being promoted.

packages/cqrs/lib/src/cqrs_result.dart
kotertom, line 1: I'm not entirely convinced to having all those separate classes for query, command and operation results. What do you think about having a single sealed class Result with subtypes Success and Failure, and then typedefs for query/command/operation results, i.e. the thing below?

```dart

typedef QueryResult<T> = Result<T, CqrsError>;

typedef CommandResult = Result<void, CqrsError>;

typedef OperationResult<T> = Result<T, CqrsError>;
```

On another note, are we planning to have generated validation errors per command?

shilangyu: > On another note, are we planning to have generated validation errors per command?

What do you mean?

> Also, I'd rather see separate error types for query, command and operation (i.e. QueryError, CommandError and OperationError instead of a common CqrsError). Queries do not have validation errors, for instance.

+1

packages/cqrs/lib/src/cqrs_result.dart
```diff
@@ -0,0 +1,140 @@
...
+// limitations under the License.
+
+import 'package:equatable/equatable.dart';
+
+import 'command_result.dart';
+import 'cqrs_error.dart';
+
+/// Generic result for CQRS query result. Can be either [QSuccess]
+/// or [QFailure].
+sealed class QResult<T> extends Equatable {
```
Albert221, line 22: Can we use full name instead of a single letter in the results? I don't like it and I think `OperationFailure` isn't _that_ much longer that `OFailure`.

When we have an if-case 3 levels deep (e.g. class -> method -> try) we have 32 characters on the line left for the variable name we are casing on in case of longest `OperationFailure`:

```dart
class A {
  void a() {
    try {
      if (a case CommandFailure(:final data)) {/*                             */
    }
  }
```

packages/cqrs/lib/src/cqrs_result.dart
kotertom: > > On another note, are we planning to have generated validation errors per command?
>
> What do you mean?

I mean that different commands have different validation errors. The frontend guild's generator creates distinct types for all validation errors, allowing them to have the compiler check whether they handled all cases. For example we can have a RegisterCommand with errors "username already taken", "password too short" etc, and another SubmitBlogPostCommand with errors "content must not be empty", "title too long". We could have a generic `CommandError<E>` class that has a list of validation errors, and those errors are a sum type with all possible variants. E.g. the `RegisterCommand` would have a matching `sealed class RegisterCommandValidationError` with variants `final class UsernameAlreadyTaken extends RegisterCommandValidationError {}`, `final class PasswordTooShort extends RegisterCommandValidationError {}` and so on. Or an enum, or something.

shilangyu: Ah ok. This is something that I already considered, to just generate the command errors as a simple enum in the contractsgenerator. It will require a little bit of work, I will mention this idea in things-to-do. Then we might even just have `Command` be generic over the error type.

packages/cqrs/lib/src/cqrs_result.dart
jtarkowski27: I agree and I'm going to change that. However, this would cause a minor conflict with already existing `CommandResult`. My solution to that is to rename the already existing `CommandResult` to `CommandResponse`. What do you think?

packages/cqrs/lib/src/cqrs_result.dart
shilangyu: > I'm not entirely convinced to having all those separate classes for query, command and operation results. What do you think about having a single sealed class Result with subtypes Success and Failure, and then typedefs for query/command/operation results, i.e. the thing below?
> 
> ```dart
> typedef QueryResult<T> = Result<T, CqrsError>;
> 
> typedef CommandResult = Result<void, CqrsError>;
> 
> typedef OperationResult<T> = Result<T, CqrsError>;
> ```

The benefit is unclear to me. Also, this made the middleware interface much more cumbersome imo.

jtarkowski27:
> The benefit is unclear to me. Also, this made the middleware interface much more cumbersome imo.

@shilangyu The main benefit is that you can map the generic result to thet of given CQRS method which makes middleware simpler since you have only one method to handle the result which imo could be quite useful. 

However while implementing this I stumbled across other problems mainly with `CommandResult`.
1. Having list of validation errors requires generic error type to be a class or structure of some sort rather than enum. Thus field `error` in `CommandFailure` is a class which has `errorType` of enum type `CommandErrorType` and list of `ValidationError` objects. This makes result classes not "homogeneous" because result classes of query and command in `Query`/`OperationFailure` have `error` field of enum type which makes things a bit less intuitive.
2. Having error of structure like type for `CommandResult` adds extra layer of nesting for example to match `validationErrors` you would have to do something like this:
```dart
if (result case CommandFailure(isInvalid: true, error: CommandError(:final validationErrors))) {...}
```
instead of simply:
```dart
if (result case CommandFailure(isInvalid: true, :final validationErrors)) {...}
```
You could always add a `validationErrors` getter with extension on `CommandFailure` more or less like this:
```dart
extension CommandFailureValidationErrorExtension on CommandResult {
  bool get isInvalid => switch (this) {
        CommandFailure(error: CommandErrorType.validation) => true,
        _ => false
      };
}
```
But this adds extra ways of getting the same instance:
```dart
failureResult.validationErrors
failureResult.error.validationErrors 
```
3. Other way is to extend generic `Result` like this: 
```dart
class CommandFailure extends Failure<void, CommandError> {
  final List<ValidationError> validationErrors;
}
```
However this way we cut ourselves off from exhaustiveness of `CommandResult`:
```dart
switch(result) {
  CommandSuccess() => ...
  CommandFailure() => ...
} // Does not exhaust [Failure<T, E>] case
```

Considering all the points above I went back to the original implementation of three separate result classes that share no custom result class.

packages/cqrs/lib/src/cqrs.dart
```diff
@@ -35,48 +60,62 @@ class Cqrs {
...
 
   /// Send a query to the backend and expect a result of the type `T`.
   ///
   /// Headers provided in `headers` are on top of the `headers` from the [Cqrs]
   /// constructor, meaning `headers` override `_headers`. `Content-Type` header
   /// will be ignored.
   ///
-  /// A [CqrsException] will be thrown in case of an error.
-  Future<T> get<T>(
+  /// After succesfull completion returns [QuerySuccess] with recieved data
```
shilangyu, line 107:
```suggestion
  /// After successful completion returns [QuerySuccess] with received data
```

packages/cqrs/lib/src/cqrs.dart
```diff
@@ -87,34 +126,20 @@ class Cqrs {
   /// constructor, meaning `headers` override `_headers`. `Content-Type` header
   /// will be ignored.
   ///
-  /// A [CqrsException] will be thrown in case of an error.
+  /// After succesfull completion returns [CommandSuccess].
```
shilangyu, line 129:
```suggestion
  /// After successful completion returns [CommandSuccess].
```

packages/cqrs/lib/src/cqrs.dart
```diff
@@ -87,34 +126,20 @@ class Cqrs {
...
+  /// A [CommandFailure] will be returned with according [CommandError]
```
shilangyu, line 130:
```suggestion
  /// A [CommandFailure] will be returned with an according [CommandError]
```

packages/cqrs/lib/src/cqrs.dart
```diff
@@ -35,48 +60,62 @@ class Cqrs {
...
   /// Send a query to the backend and expect a result of the type `T`.
   ///
   /// Headers provided in `headers` are on top of the `headers` from the [Cqrs]
   /// constructor, meaning `headers` override `_headers`. `Content-Type` header
   /// will be ignored.
   ///
-  /// A [CqrsException] will be thrown in case of an error.
-  Future<T> get<T>(
+  /// After succesfull completion returns [QuerySuccess] with recieved data
+  /// of type `T`. A [QueryFailure] will be returned with according
```
shilangyu, line 108:
```suggestion
  /// of type `T`. A [QueryFailure] will be returned with an according
```

packages/cqrs/lib/src/cqrs.dart
```diff
@@ -124,31 +149,151 @@ class Cqrs {
   /// constructor, meaning `headers` override `_headers`. `Content-Type` header
   /// will be ignored.
   ///
-  /// A [CqrsException] will be thrown in case of an error.
-  Future<T> perform<T>(
+  /// After succesfull completion returns [OperationSuccess] with recieved
```
shilangyu, line 152:
```suggestion
  /// After successful completion returns [OperationSuccess] with received
```

packages/cqrs/lib/src/cqrs.dart
```diff
@@ -124,31 +149,151 @@ class Cqrs {
...
+  /// data of type `T`. A [OperationFailure] will be returned with
+  /// according [OperationError] in case of an error.
```
shilangyu, line 154:
```suggestion
  /// an according [OperationError] in case of an error.
```

shilangyu (approved):
Typos and one question, otherwise LGTM

packages/cqrs/lib/src/cqrs.dart
jtarkowski27: Yes, I reverted this by mistake

kotertom (approved):

packages/cqrs/CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
+# 10.0.0
+
+- **Breaking:** Fundamentaly change overall `Cqrs` API making it no-throw guarantee.
+- **Breaking:** Make `Cqrs.get`, `Cqrs.run` and `Cqrs.perform` return result data in form of `QueryResult`, `CommandResult` and `OperationResult` respectively.
+- Add `logger` parameter (of type `Logger` from `logging` package) to `Cqrs` default constructor. If provided, the `logger` will be used as a debug logging interface in execution of CQRS methods.
+- Add middleware mechanism in form of `CqrsMiddleware` intended to use in processing result from queries, commands and operations.
+- **Breaking:** Remove `CqrsException`.
+- **Breaking:** Rename previous `CommandResult` to `CommandResponse` and make it package private.
+- Mark the `CqrsMethod` as `sealed`.
+- **Breaking:** Make `ValidationError` extend `Equatable` from `equatable` package.
```
shilangyu, line 10: Is that breaking?

packages/cqrs/CHANGELOG.md
```diff
@@ -1,3 +1,15 @@
...
+- **Breaking:** Add `equatable` (`2.0.5`) and `logging` (`1.2.0`) dependencies.
```
shilangyu, line 11:
```suggestion
- **Breaking:** Add `equatable` (`^2.0.5`) and `logging` (`^1.2.0`) dependencies.
```

shilangyu (approved):

packages/cqrs/CHANGELOG.md
jtarkowski27: I'm not sure, I'd leave it just in case

PR #161, opened 2023-09-27
Fix cqrs v10.0.0 issues

jtarkowski27:
- Rename `forbiddenAccess` to `authorization` in error enums.
- Update error enums docs.
- Fix `CqrsMiddleware` not being exported from the package.
- Move `hasError` and `hasErrorForProperty` from `CommandResponse` to `CommandFailure`.

### DIFF
--- a/packages/cqrs/CHANGELOG.md
+++ b/packages/cqrs/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 10.0.1
+
+- **Breaking:** Rename `forbiddenAccess` to `authorization` in error enums.
+- Update error enums docs.
+- Fix `CqrsMiddleware` not being exported from the package.
+
 # 10.0.0
 
 - **Breaking:** Fundamentaly change overall `Cqrs` API making it no-throw guarantee.
--- a/packages/cqrs/CHANGELOG.md
+++ b/packages/cqrs/CHANGELOG.md
@@ -3,6 +3,7 @@
 - **Breaking:** Rename `forbiddenAccess` to `authorization` in error enums.
 - Update error enums docs.
 - Fix `CqrsMiddleware` not being exported from the package.
+- Move `hasError` and `hasErrorForProperty` from `CommandResponse` to `CommandFailure`.
 
 # 10.0.0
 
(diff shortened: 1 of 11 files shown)


packages/cqrs/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# 10.0.1
+
+- **Breaking:** Rename `forbiddenAccess` to `authorization` in error enums.
```
shilangyu, line 3: I'll let @Albert221 decide what to do here. Clearly its not a patch release as it is breaking. But I understand the sentiment that this is a hotfix over v10. I don't think pub has any process of yanking releases.

Albert221: I'll mark 10.0.0 as retracted in Pub. Please mark 10.0.0 as retracted also in changelog @jtarkowski27

Albert221:
Regarding errors (network, unknown), I think it would make sense to also expose the original error/exception. What do you think? @shilangyu @jtarkowski27

packages/cqrs/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+- Update error enums docs.
+- Fix `CqrsMiddleware` not being exported from the package.
+- Move `hasError` and `hasErrorForProperty` from `CommandResponse` to `CommandFailure`.
+
 # 10.0.0
```
Albert221, line 9:
```suggestion

**Retracted.**

```

packages/cqrs/CHANGELOG.md
shilangyu: Good to know there is some yanking support in pub

shilangyu:
> Regarding errors (network, unknown), I think it would make sense to also expose the original error/exception. What do you think? @shilangyu @jtarkowski27

Probably not. Our cqrs is quite contained in the errors it can exhibit. If you get an unknown error nothing much can be done with the exception/stacktrace anyways, other than to log it which is already being done.

shilangyu (approved):

PR #194, opened 2023-10-16
Migrate debug page to leancode_lint

shilangyu:


packages/debug_page/test/logging_http_client_test.dart
```diff
@@ -40,9 +40,6 @@ void main() {
             statusCode,
             contentLength: contentLength,
             request: request,
-            headers: headers,
-            isRedirect: isRedirect,
-            persistentConnection: persistentConnection,
```
gogolon, line 45: why?

gogolon: if someone changed the values in the test (let's say `isRedirect = true`)  the test would fail even if the feature worked well. Also I did not figure out why it is that way at first glance (but this could be fixed be prefixing constants in test with `default`, e.g `defaultIsRedirect`).  Not a big issue for me but I think this could be a little bit misleading

gogolon (approved):

packages/debug_page/test/logging_http_client_test.dart
shilangyu: reverting, I agree with you

PR #195, opened 2023-10-16
Update dev dependencies in packages

shilangyu:


.github/workflows/enhanced_gradients-publish.yml
```diff
@@ -22,15 +22,10 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Set up Dart
-        uses: dart-lang/setup-dart@v1
-        with:
-          sdk: 3.0.0
-
       - name: Set up Flutter
```
shilangyu, line 25: @bartekpacia do you know if this was needed? I am pretty sure that installing flutter installs dart as well

.github/workflows/override_api_endpoint-test.yml
```diff
@@ -20,8 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - version: '2.5.0'
-          - version: '3.3'
+          - version: '3.13.7'
```
bartekpacia, line 23: I prefer having the newest PATCH e.g. `3.13.x`

Unfortunately it doesn't work for `dart-lang/setup-dart` action.

bartekpacia (approved):

.github/workflows/enhanced_gradients-publish.yml
bartekpacia: TL;DR publishing doesn't work without it

See https://github.com/leancodepl/mobile-tools/blob/0a67ef6c49933a560fb3f6633465ce83032dfdf8/.github/actions/pub-release/action.yaml#L74-L76 for more context.

.github/workflows/override_api_endpoint-test.yml
shilangyu: I think it works now https://github.com/dart-lang/setup-dart/pull/106/files

bartekpacia: oh wow that's awesome

Albert221 (approved):

PR #201, opened 2023-10-25
Add compute implementation from `flutter_svg` that does not spawn isolates in debug mode

jtarkowski27:
The fact that `compute` spawns isolates in debug mode caused massive drops in performance related to `.svg` assets loading. When I had a tens of assets in the same view, it took ~5s to load all of them using `FlutterSvgAdaptiveLoader`. That was approx 10x worse than using `SvgAssetLoader` from `flutter_svg`. 

For that reason I brought more optimised `compute` implementation for debug mode from `flutter_svg` package. It seems that this version of `compute` solves performance issues.

### DIFF
--- a/packages/leancode_flutter_svg_adaptive_loader/lib/src/compute.dart
+++ b/packages/leancode_flutter_svg_adaptive_loader/lib/src/compute.dart
@@ -31,10 +31,6 @@ Future<R> _testCompute<Q, R>(
   Q message, {
   String? debugLabel,
 }) {
-  if (foundation.kDebugMode) {
-    final bindingType = foundation.BindingBase.debugBindingType();
-    if (bindingType.toString() == 'AutomatedTestWidgetsFlutterBinding') {}
-  }
   final result = callback(message);
   if (result is Future<R>) {
     return result;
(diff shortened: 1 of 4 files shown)


packages/leancode_flutter_svg_adaptive_loader/lib/src/compute.dart
```diff
@@ -0,0 +1,49 @@
...
+import 'package:flutter/foundation.dart' as foundation;
+
+Future<R> _testCompute<Q, R>(
+  foundation.ComputeCallback<Q, R> callback,
+  Q message, {
+  String? debugLabel,
+}) {
+  if (foundation.kDebugMode) {
+    final bindingType = foundation.BindingBase.debugBindingType();
+    if (bindingType.toString() == 'AutomatedTestWidgetsFlutterBinding') {}
```
Albert221, line 36: It's dead code, it's not used

Albert221 (approved):

shilangyu:
What about release mode? Is the performance problem visible?

jtarkowski27:
> What about release mode? Is the performance problem visible?

@shilangyu On release performance is almost the same as with `SvgAssetLoader` so the issue only affects debug mode.

PR #203, opened 2023-10-30
Change the default behaviour of force update, change the way of displaying suggest update dialogs

gogolon:
The decision has been made to alter the default behaviour of the package. 
By default, VersionSupport responses will now be applied immediately once obtained, instead of waiting till next app launch. Former behaviour can be restored using flags.
This PR also changes the way of displaying suggest update dialog to make it fully independent of code in the project where the package is used.


packages/force_update/lib/src/force_update_storage.dart
```diff
@@ -7,20 +7,20 @@ import 'package:version/version.dart';
...
     required this.versionAtTimeOfRequest,
-    required this.result,
+    required this.conclusion,
   });
 
   ForceUpdateResult.fromJson(Map<String, dynamic> json)
       : versionAtTimeOfRequest =
             Version.parse(json['versionAtTimeOfRequest'] as String),
-        result = VersionSupportResultDTO.values[json['result'] as int];
+        conclusion = VersionSupportResultDTO.values[json['conclusion'] as int];
```
gogolon, line 16: this is somewhat breaking, but the former name does not make sense, it caused some weird expressions like `forceUpdateResult.result`. So I am convinced that this should be changed, but I am not sure how. Should I handle reading json differently if there is a result stored by a previous version of the package? Have we even used this package anywhere yet?

packages/force_update/README.md
```diff
@@ -3,12 +3,9 @@
 This is an internal Leancode package for quickly implementing Force Update. To use it, just wrap your `MaterialApp` in a `ForceUpdateGuard`
 widget and pass an instance of CQRS.
 
-# Behavior
+# Behaviour
```
PiotrRogulski, line 6: Use American spelling

packages/force_update/README.md
```diff
@@ -3,12 +3,9 @@
...
-To keep the minimum required version up to date, requests are made every 5 minutes when the app is running.
+On every app launch, a request will be made to fetch the minimum required version of the app. By default, the result will be applied immediately after the response is obtained - if necessary, the package will display either force update screen or a dialog suggesting to update the app. If you do not want to interrupt the user after they have launched the app (for instance, they could have launched it without internet connection and then turn it on at some point) and would rather only show a force update screen or a dialog on app launch, you can change the behaviour of the package to make it store the responses and apply them on subsequent app launches. This can be done by setting `showForceUpdateScreenImmediately` or `showSuggestUpdateDialogImmediately` to false in `ForceUpdateGuard`. Regardless of those settings, to keep the minimum required version up to date, requests are made every 5 minutes while the app is running.
```
PiotrRogulski, line 8: Use en dashes instead of hyphens when appropriate

packages/force_update/example/lib/main.dart
```diff
@@ -23,39 +23,37 @@ void main() async {
...
     super.key,
-    required ForceUpdate forceUpdate,
+    required ForceUpdateController forceUpdateController,
     required Cqrs cqrs,
-  })  : _forceUpdate = forceUpdate,
+  })  : _forceUpdate = forceUpdateController,
         _cqrs = cqrs;
 
-  final ForceUpdate _forceUpdate;
+  final ForceUpdateController _forceUpdate;
```
PiotrRogulski, line 45:
```suggestion
  final ForceUpdateController _forceUpdateController;
```

packages/force_update/README.md
gogolon: why? do we have such convention? I changed this for unification

PiotrRogulski: I think that we usually use American spelling, from the projects I've seen

packages/force_update/lib/src/force_update_guard.dart
```diff
@@ -56,51 +70,57 @@ class _ForceUpdateGuardState extends State<ForceUpdateGuard> {
...
       }
+    }
+  }
 
-      final context = widget.dialogContextKey.currentContext;
+  Future<void> _updateAndMaybeApplyVersionsInfo() async {
+    await _updateVersionsInfo();
+
+    final recentResult = await _storage.readMostRecentResult();
+    return switch ((
```
RobertOdrowaz, line 99: IMO normal if would be more readable here

RobertOdrowaz (approved):

packages/force_update/README.md
gogolon: does "when appropriate" mean every time in this case?

PiotrRogulski: Yes, it would seem so 😄

packages/force_update/lib/src/force_update_guard.dart
gogolon: I thought the same until I tried it - the names are very long and I'm not sure if it actually helps
```
    if ((recentResult?.conclusion == VersionSupportResultDTO.updateRequired &&
            widget.showForceUpdateScreenImmediately) ||
        (recentResult?.conclusion == VersionSupportResultDTO.updateSuggested &&
            widget.showSuggestUpdateDialogImmediately)) {
      return _applyResult(result: recentResult);
    }
```

PiotrRogulski (approved):

PR #204, opened 2023-10-30
Create cache key early to avoid resolving DefaultAssetBundle when ancestor is deactivated

jtarkowski27:
Former implementation of `FlutterSvgAdaptiveLoader.loadBytes` causes exception:

```
FlutterError (Looking up a deactivated widget's ancestor is unsafe.
At this point the state of the widget's element tree is no longer stable.
To safely refer to a widget's ancestor in its dispose() method, save a reference to the ancestor by calling dependOnInheritedWidgetOfExactType() in the widget's didChangeDependencies() method.)
```

### DIFF
--- a/packages/leancode_flutter_svg_adaptive_loader/CHANGELOG.md
+++ b/packages/leancode_flutter_svg_adaptive_loader/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.1.1
 
-- Fix error caused by resolving DefaultAssetBundle
+- Fix error caused by resolving `DefaultAssetBundle` from context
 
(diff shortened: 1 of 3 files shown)


packages/leancode_flutter_svg_adaptive_loader/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## 1.1.0
```
Albert221, line 1: This would be 1.1.1

Albert221 (approved):

Xezolpl (approved):

PR #235, opened 2024-01-29
Add avoid_single_child_in_multi_child_widgets lint rule

shilangyu:


packages/leancode_lint/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 8.1.0
+
+- Enable custom lint `avoid_single_child_in_multi_child_widgets` for `Column`, `Row`, `Flex`, and `Wrap`
```
shilangyu, line 3: Feel free to suggest other ones, I can easily add them

PiotrRogulski: Maybe also `Stack`? Usually could be replaced by `Align` or stuff

packages/leancode_lint/test/lints_test_app/lib/avoid_single_child_in_multi_child_widget_test.dart
```diff
@@ -0,0 +1,194 @@
...
+    );
+  }
+}
+
+class ZeroChildrenIsNotMarked extends StatelessWidget {
+  const ZeroChildrenIsNotMarked({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return const Wrap();
```
PiotrRogulski, line 67: Could be a separate lint, but maybe we want to disallow no children as well?

packages/leancode_lint/CHANGELOG.md
PiotrRogulski: We could also detect equivalent sliver widgets — `SliverMainAxisGroup` and `SliverCrossAxisGroup`

packages/leancode_lint/test/lints_test_app/lib/avoid_single_child_in_multi_child_widget_test.dart
shilangyu: I thought about it, but figured that there are so many widgets that dont make sense when used with no children/child. Also I would assume it happens way less often, while single child in Column/Row happens quite often by accident. So yeah, if we want a no children/child lint it should be separate

PR #236, opened 2024-01-29
Use mobile-tools for releasing leancode_lint

### LINKED ISSUE #234: Use mobile-tools for releasing `leancode_lint`
> todo: leancode_lint does not use mobile_tools to publish to the pub, so it does not create appropriate tags in the repo

_Originally posted by @Albert221 on Slack_

KrzysztofMamak:
Closes #234

### DIFF
--- a/.github/workflows/leancode_lint-publish.yml
+++ b/.github/workflows/leancode_lint-publish.yml
@@ -21,10 +21,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Setup Dart
+      - name: Set up Dart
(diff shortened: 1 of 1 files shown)


RobertOdrowaz (approved):

shilangyu (commented):
I don't remember what the problem was exactly, but i remember that this script would not work without setting up flutter in the CI. @Albert221 do you recall anything like that?

Albert221:
@shilangyu we need to install Flutter on CI so that pub get done by pub publish won't complain about missing Flutter SDK if it's in the deps

shilangyu (approved):

PR #242, opened 2024-02-15
Disable deprecated 3.3.0 lints

PiotrRogulski:
Disables a few lints that were deprecated or removed for the 3.3.0 Dart release. Also bumps the SDK to that version.

### DIFF
--- a/packages/leancode_lint/CHANGELOG.md
+++ b/packages/leancode_lint/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 9.0.0
+- Disable lints deprecated in Dart 3.3.0
+- Bump minimum Dart version to 3.3
+
 # 8.1.0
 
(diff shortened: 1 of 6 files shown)


shilangyu (changes requested):
Add changelog entry. You can bump major

packages/leancode_lint/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 9.0.0
+- Disable lints deprecated in Dart 3.3.0
+- Bump minimum Dart version to 3.3
+
 # 8.1.0
```
shilangyu, line 5: This was never released actually, and it should've been in a major anyway. Could you merge those changelogs please?

shilangyu (changes requested):

packages/leancode_lint/CHANGELOG.md
PiotrRogulski: Yup :+1:

shilangyu (approved):

PR #249, opened 2024-03-07
Add leancode markup

pdenert:


shilangyu (approved):
TL;DR but the commits look correctly ported

pdenert:
> Do we need all these platforms in `example/`?

We support all of them, so why not?

packages/leancode_markup/lib/src/parser/markup_parser.dart
```diff
@@ -0,0 +1,108 @@
...
+            'Found a closing tag without a matching opening one.',
+            source,
+          );
+        } else {
+          tagStack.removeLast();
+        }
+      case TextToken(:final content):
+        yield TaggedText(
+          content,
+          tags: [...tagStack],
```
Albert221, line 88: Do we want to have a flat structure of text spans instead of using nested spans?

| What we have now | What I though we want to have |
|-|-|
| `[b]aaa [i]bbb[/i][/b]` | `[b]aaa [i]bbb[/i][/b]` | 
| `TextSpan(b, aaa) + TextSpan(b+i, bbb)` | `TextSpan(b, aaa + TextSpan(i, bbb))`

pdenert: I have 2 thoughts and one question. 
1. What's the profit from the second option?
2. This PR is about to move one repo into other, so it would be better to change this comment into issue.
3. I'm not sure, but there may be a problem with tag factories in the nested spans.

Albert221: 1. We rely on Flutter's `TextStyle` merging instead of our own. Plus it may be helpful when we'll for example add some kind of a stateful gesture recognizer (for `[url]`) that will underline the clicked/hovered text span.

Albert221: Issue #250

PR #261, opened 2024-03-22
Add linter rule to disallow string literals in `Widget` classes

### LINKED ISSUE #260: Add custom linter rule to prevent usage of string literals in widgets
When using internalization (e.g. through a package such as `i18n`), no text placed in the UI should be hardcoded to one language. We could implement a custom linter rule to prevent this from happening.

bdmendes:
Closes #260 
This linter rule makes code review easier for internationalized apps, by preventing the usage of string literals in widgets.

This may be annoying, since we need string literals for things other than UI inside a widget's class, and it's not possible (at least from what I've seen from the analyzer API) to guarantee with a 100% certainty that they are being used in that context. As such, I added 2 simple exceptions:
- When handling an exception: so that we can log it/send to a server
- When throwing an exception: so that we can define a proper message

Furthermore, there are some things we should discuss:
- Should we make these ancestor exceptions configurable?
- Should the exceptions be related instead to e.g. the number of the function that is being invoked?

### DIFF
--- a/packages/leancode_lint/lib/leancode_lint.dart
+++ b/packages/leancode_lint/lib/leancode_lint.dart
@@ -5,6 +5,7 @@ import 'package:leancode_lint/assists/convert_record_into_nominal_type.dart';
 import 'package:leancode_lint/lints/add_cubit_suffix_for_cubits.dart';
 import 'package:leancode_lint/lints/avoid_conditional_hooks.dart';
 import 'package:leancode_lint/lints/avoid_single_child_in_multi_child_widget.dart';
+import 'package:leancode_lint/lints/avoid_string_literals_in_widgets.dart';
 import 'package:leancode_lint/lints/catch_parameter_names.dart';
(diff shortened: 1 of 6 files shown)


shilangyu:
Did you try to enable it in some project to see what kind of false positives you got?

Generally speaking, I think I would for approaching this more conservatively. As I said, we'd rather miss some violations than annoy the user with false-positives. I was thinking that maybe we should just complain if a string literal is used directly in a widget constructor, as I think this is actually 90% of the cases (source: I made it up).

Then, it could be configurable what kind of constructor parameters should not be flagged (for instance `TextField.obscuringCharacter` seems like something that shouldn't be flagged). Also, maybe the empty string literal shouldn't be ever flagged, its probably used as some UI workaround.

What do you think?

bdmendes:
Hi! Sorry for letting this go unnoticed. I'm no longer working with Flutter regularly, so I would appreciate if you could take over this for me. Anyway, I'd agree that we can be more conservative here.

PR #286, opened 2024-04-27
Treat HookWidget as a new hook context

### LINKED ISSUE #284: [leancode_lint] `avoid_conditional_hooks` false positive with `HookBuilder` inside `HookWidget`
`avoid_conditional_hooks` rule reports a false positive if there is conditional `HookBuilder` inside a `HookWidget`. If the containing widget is a `StatelessWidget` everything works fine.

image

```dart
class HookBuilderInIfShouldBeOk extends HookWidget {
  const HookBuilderInIfShouldBeOk({super.key});

  @override
  Widget build(BuildContext context) {
    if (Random().nextBool()) {
      return const SizedBox();
    }

    return HookBuilder(
      builder: (context) {
        useState(1);
        return const SizedBox();
      },
    );
  }
}
```

shilangyu:
closes #284

### DIFF
--- a/packages/leancode_lint/test/lints_test_app/pubspec.lock
+++ b/packages/leancode_lint/test/lints_test_app/pubspec.lock
@@ -236,7 +236,7 @@ packages:
       path: "../.."
       relative: true
     source: path
-    version: "12.0.0"
+    version: "12.1.0"
   logging:
     dependency: transitive
     description:
--- a/packages/leancode_lint/test/lints_test_app/pubspec.lock
+++ b/packages/leancode_lint/test/lints_test_app/pubspec.lock
@@ -236,7 +236,7 @@ packages:
       path: "../.."
       relative: true
     source: path
-    version: "11.0.0"
+    version: "12.0.0"
   logging:
     dependency: transitive
     description:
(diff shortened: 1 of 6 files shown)


packages/leancode_lint/test/lints_test_app/pubspec.lock
```diff
@@ -257,18 +257,18 @@ packages:
     dependency: transitive
     description:
       name: material_color_utilities
-      sha256: f7142bb1154231d7ea5f96bc7bde4bda2a0945d2806bb11670e30b850d56bdec
+      sha256: "0e0a020085b65b6083975e499759762399b4475f766c21668c4ecca34ea74e5a"
       url: "https://pub.dev"
     source: hosted
-    version: "0.11.1"
+    version: "0.8.0"
```
PiotrRogulski, line 263: Downgrade 🤨

PiotrRogulski (approved):

packages/leancode_lint/test/lints_test_app/pubspec.lock
shilangyu: Unclear to me why that is. It tells me for example meta is pinned by flutter at 1.11.0, so no clue how come it was 1.12.0 before

PiotrRogulski: It might be that the example app doesn't have a specific Flutter version specified, and you ran it with an older version :thinking:

shilangyu: image
 🤷

PiotrRogulski: Let's keep it downgraded, then :thinking:

PR #321, opened 2024-07-01
[leancode_debug_page] Avoid accelerometer exception when app is run on a device which does not support it 

gogolon:


packages/leancode_debug_page/lib/src/core/debug_page_controller.dart
```diff
@@ -9,6 +9,10 @@ import 'package:logging/logging.dart';
 import 'package:rxdart/rxdart.dart';
 import 'package:shake/shake.dart';
 
+bool get _isMobile =>
```
RobertOdrowaz, line 12: I think `Platform.*` may be more correct in this case

gogolon: why?

PiotrRogulski: I wouldn't use `Platform`; that will effectively prevent apps from running on the web

RobertOdrowaz: It can be `Platform` from `universal_io`. This is exactly the case described [here](https://api.flutter.dev/flutter/foundation/defaultTargetPlatform.html)
> The [dart:io.Platform](https://api.flutter.dev/flutter/dart-io/Platform-class.html) object should only be used directly when it's critical to actually know the current platform, without any overrides possible (for example, when a system API is about to be called)

RobertOdrowaz (approved):

PR #325, opened 2024-07-04
[leancode_debug_page] Vendor shake package, bump sensors_plus dependency

gogolon:


packages/leancode_debug_page/lib/src/core/shake_detector.dart
```diff
@@ -0,0 +1,80 @@
...
+// Shake dependency can be brought back once https://github.com/deven98/shake/pull/32
+// is merged
+
+import 'dart:async';
+import 'dart:math';
+
+import 'package:flutter/foundation.dart';
+import 'package:sensors_plus/sensors_plus.dart';
+
+class ShakeDetector {
```
RobertOdrowaz, line 13: We need to include `shake` in our LICENSE file in this case

RobertOdrowaz (approved):

PR #328, opened 2024-07-23
Change diagnostic location for `constructor_parameters_and_fields_should_have_the_same_order`

karolinaoparczyk:
Changes how `constructor_parameters_and_fields_should_have_the_same_order` reports issues.

Right now the issue is reported just above the first parameter, which doesn't always cause the problem. In my opinion it would be better if the lint reported issues above the constructor.
This is especially confusing when one wants to use `ignore`:

How it looks now:
```
class MyClass{
    MyClass({
        // ignore: constructor_parameters_and_fields_should_have_the_same_order
        required this.value,
        required this.label,
        required this.enabled,
    });

    final bool enabled;
    final double value;
    final String label;
}
```

How it looks after my changes:
```
class MyClass{
    // ignore: constructor_parameters_and_fields_should_have_the_same_order
    MyClass({
        required this.value,
        required this.label,
        required this.enabled,
    });

    final bool enabled;
    final double value;
    final String label;
}
```

### DIFF
--- a/packages/leancode_lint/lib/lints/constructor_parameters_and_fields_should_have_the_same_order.dart
+++ b/packages/leancode_lint/lib/lints/constructor_parameters_and_fields_should_have_the_same_order.dart
@@ -34,15 +34,9 @@ class ConstructorParametersAndFieldsShouldHaveTheSameOrder
         final constructors = node.members.whereType<ConstructorDeclaration>();
         for (final constructor in constructors) {
           if (!_hasValidOrder(constructor, fields)) {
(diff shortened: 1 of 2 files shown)


gogolon (approved):

PiotrRogulski (approved):

shilangyu (approved):

shilangyu (changes requested):
Please add a changelog entry

shilangyu (approved):

shilangyu:
Can we merge this?

PR #338, opened 2024-08-22
Always set `tokenEndpoint` using value of `authorizationUri` from `OAuthSettings` during `LoginClient` initialization.

### LINKED ISSUE #334: Avoid using stored tokenEndpoint for credentials refresh to prevent authentication issues when auth server URL changes
Currently `login_client_flutter` saves `Credentials` object that contains [tokenEndpoint](https://github.com/leancodepl/flutter_corelibrary/blob/master/packages/login_client_flutter/lib/src/flutter_secure_credentials_storage.dart#L34) that may change in time (eg. auth server url may change).

Should we assume that we want to use in order to authenticate the user when the access token expires? Currently we use `tokenEndpoint` that is stored within the app keychain. This means that if someone is logged in to the app, they won't be able to refresh their credentials until the application won't clear the stored `Credentials`. 

**Proposed solution:**

If we were to refresh credentials we should use auth server endpoint defined in [here](https://github.com/leancodepl/flutter_corelibrary/blob/master/packages/login_client/lib/src/oauth_settings.dart#L19), not the one we store in the keychain.

pdenert:
fix #334

### DIFF
--- a/packages/login_client/CHANGELOG.md
+++ b/packages/login_client/CHANGELOG.md
@@ -1,5 +1,6 @@
-# Unreleased
+# 3.2.0
 
+- Add option to override `tokenEndpoint` from saved `Credentials` during initialization.
 - Bump `leancode_lint` dev dependency to `12.0.0`.
 - Bump `custom_lint` dev dependency to `0.6.4`.
 
--- a/packages/login_client/CHANGELOG.md
+++ b/packages/login_client/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 3.2.0
 
-- Add option to override `tokenEndpoint` from saved `Credentials` during initialization.
+- Always set `tokenEndpoint` using value of `authorizationUri` from `OAuthSettings` during `LoginClient` initialization.
 - Bump `leancode_lint` dev dependency to `12.0.0`.
 - Bump `custom_lint` dev dependency to `0.6.4`.
 
--- a/packages/login_client/CHANGELOG.md
+++ b/packages/login_client/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 3.2.0
 
-- Always set `tokenEndpoint` using value of `authorizationUri` from `OAuthSettings` during `LoginClient` initialization.
+- Always use `OAuthSettings.authorizationUri` for token endpoint in `LoginClient` initialization (for refreshing the token).
 - Bump `leancode_lint` dev dependency to `12.0.0`.
 - Bump `custom_lint` dev dependency to `0.6.4`.
 
--- a/packages/login_client/lib/src/login_client.dart
+++ b/packages/login_client/lib/src/login_client.dart
@@ -107,8 +107,8 @@ class LoginClient extends http.BaseClient {
     }
   }
 
-  /// Restores saved credentials from the credentials storage and set
-  /// `tokenEndpoint`to the one provided in the `oAuthSettings`.
+  /// Restores saved credentials from the credentials storage and sets
+  /// `tokenEndpoint` to the one provided in the `oAuthSettings`.
   Future<Credentials?> get _credentialsToInitialize async {
     final credentials = await _credentialsStorage.read();
 
--- a/packages/login_client/lib/src/login_client.dart
+++ b/packages/login_client/lib/src/login_client.dart
@@ -84,10 +84,9 @@ class LoginClient extends http.BaseClient {
 
   /// Restores saved credentials from the credentials storage.
   ///
-  /// You can provide a custom `tokenEndpoint` to override the one stored in
-  /// the credentials. This is useful when the token endpoint changes.
+  /// `_oAuthSettings.authorizationUri` is used as the `tokenEndpoint`.
   Future<void> initialize() async {
-    final credentials = await _credentialsToInitialize;
+    final credentials = await getCredentialsToInitialize();
 
     if (credentials != null) {
       _oAuthClient = buildOAuth2ClientFromCredentials(
@@ -109,7 +108,7 @@ class LoginClient extends http.BaseClient {
 
   /// Restores saved credentials from the credentials storage and sets
   /// `tokenEndpoint` to the one provided in the `oAuthSettings`.
-  Future<Credentials?> get _credentialsToInitialize async {
+  Future<Credentials?> getCredentialsToInitialize() async {
     final credentials = await _credentialsStorage.read();
 
     if (credentials != null) {
--- a/packages/login_client/lib/src/login_client.dart
+++ b/packages/login_client/lib/src/login_client.dart
@@ -87,20 +87,10 @@ class LoginClient extends http.BaseClient {
   /// You can provide a custom `tokenEndpoint` to override the one stored in
   /// the credentials. This is useful when the token endpoint changes.
   Future<void> initialize({Uri? tokenEndpoint}) async {
-    var credentials = await _credentialsStorage.read();
+    final credentials =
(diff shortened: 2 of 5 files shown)


packages/login_client/lib/src/login_client.dart
```diff
@@ -87,9 +88,21 @@ class LoginClient extends http.BaseClient {
   Future<oauth2.Credentials?> get credentials => _credentialsStorage.read();
 
   /// Restores saved credentials from the credentials storage.
-  Future<void> initialize() async {
-    final credentials = await _credentialsStorage.read();
+  Future<void> initialize({Uri? tokenEndpoint}) async {
```
mkucharski17, line 86: Pls add dart doc about `tokenEndpoint`

packages/login_client/lib/src/login_client.dart
```diff
@@ -87,9 +88,21 @@ class LoginClient extends http.BaseClient {
...
+    var credentials = await _credentialsStorage.read();
+
     if (credentials != null) {
+      if (tokenEndpoint != null) {
+        credentials = Credentials(
```
mkucharski17, line 91: Doesn't Credentials class have `copyWith`?

packages/login_client/lib/src/login_client.dart
```diff
@@ -87,9 +88,21 @@ class LoginClient extends http.BaseClient {
   Future<oauth2.Credentials?> get credentials => _credentialsStorage.read();
 
   /// Restores saved credentials from the credentials storage.
-  Future<void> initialize() async {
-    final credentials = await _credentialsStorage.read();
+  Future<void> initialize({Uri? tokenEndpoint}) async {
+    var credentials = await _credentialsStorage.read();
```
mkucharski17, line 87: I dont like `var`s I would use getter for creating credential instance or using the one read from storage. that's just a suggestion.

packages/login_client/lib/src/login_client.dart
pdenert: Unfortunately, no: https://pub.dev/documentation/oauth2/latest/oauth2/Credentials-class.html

mkucharski17 (approved):

packages/login_client/lib/src/login_client.dart
```diff
@@ -87,8 +83,13 @@ class LoginClient extends http.BaseClient {
   Future<oauth2.Credentials?> get credentials => _credentialsStorage.read();
 
   /// Restores saved credentials from the credentials storage.
-  Future<void> initialize() async {
-    final credentials = await _credentialsStorage.read();
+  ///
+  /// You can provide a custom `tokenEndpoint` to override the one stored in
+  /// the credentials. This is useful when the token endpoint changes.
+  Future<void> initialize({Uri? tokenEndpoint}) async {
```
Albert221, line 89: I wouldn't accept the `tokenEndpoint` from the outside but rather a flag that would change the source of the token endpoint for refreshing it or I would change the behavior altogether (without a flag). The `tokenEndpoint` used shall come from `_oAuthSettings.authorizationUri`.

@lukaszgarstecki is it safe to make such a change to our client, that always the endpoint passed to the OAuth client settings is used for refreshing the token instead of trying to use the same endpoint that was used for the initial authorization? From the top of my head, I can't think of circumstances that would make it an unwanted behavior.

packages/login_client/lib/src/login_client.dart
```diff
@@ -107,6 +108,27 @@ class LoginClient extends http.BaseClient {
     }
   }
 
+  Future<Credentials?> _getCredentialsToInitialize({Uri? tokenEndpoint}) async {
+    final credentials = await _credentialsStorage.read();
+
+    if (credentials != null) {
+      if (tokenEndpoint != null) {
+        return Credentials(
```
Albert221, line 116: Let's create a `copyWithTokenEndpoint` private extension on `Credentials`

packages/login_client/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-# Unreleased
+# 3.2.0
 
+- Always set `tokenEndpoint` using value of `authorizationUri` from `OAuthSettings` during `LoginClient` initialization.
```
Albert221, line 3:
```suggestion
- Always use `OAuthSettings.authorizationUri` for token endpoint in `LoginClient` initialization (for refreshing the token).
```

packages/login_client/lib/src/login_client.dart
pdenert: After discussion we decided to go with the second approach (without flag). By default we always set for `tokenEndpoint` value taken from `OAuthSettings.authorizationUri`

packages/login_client/lib/src/login_client.dart
```diff
@@ -87,8 +83,12 @@ class LoginClient extends http.BaseClient {
   Future<oauth2.Credentials?> get credentials => _credentialsStorage.read();
 
   /// Restores saved credentials from the credentials storage.
+  ///
+  /// You can provide a custom `tokenEndpoint` to override the one stored in
```
Albert221, line 87: This dartdoc is no longer valid. We may mention using `oAuthSettings` for the initialization (the `refresh()` method will still use the endpoint from the `Credentials`, but that's inside the `oauth2` package and there's no nice way around that.

packages/login_client/lib/src/login_client.dart
```diff
@@ -107,6 +107,25 @@ class LoginClient extends http.BaseClient {
     }
   }
 
+  /// Restores saved credentials from the credentials storage and set
+  /// `tokenEndpoint`to the one provided in the `oAuthSettings`.
```
Albert221, line 111:
```suggestion
  /// Restores saved credentials from the credentials storage and sets
  /// `tokenEndpoint` to the one provided in the `oAuthSettings`.
```

packages/login_client/lib/src/login_client.dart
```diff
@@ -107,6 +107,25 @@ class LoginClient extends http.BaseClient {
...
+  Future<Credentials?> get _credentialsToInitialize async {
```
Albert221, line 112: We are performing some actions here (call the credentials storage), I'd prefer to make it a method to be more explicit that it hides _some_ complexity

packages/login_client/lib/src/login_client.dart
```diff
@@ -87,8 +83,12 @@ class LoginClient extends http.BaseClient {
   Future<oauth2.Credentials?> get credentials => _credentialsStorage.read();
 
   /// Restores saved credentials from the credentials storage.
+  ///
+  /// You can provide a custom `tokenEndpoint` to override the one stored in
+  /// the credentials. This is useful when the token endpoint changes.
   Future<void> initialize() async {
```
Albert221, line 89: Let's add test that checks whether the credentials do have updated `tokenEndpoint` on initialization

packages/login_client/test/login_client_test.dart
```diff
@@ -73,6 +79,39 @@ void main() {
...
+          tokenEndpoint: Uri.parse('https://example.com'),
+        );
+
+        when(() => credentialsStorage.read())
+            .thenAnswer((_) async => credentials);
+
+        loginClient.onCredentialsChanged.listen(
+          expectAsync1((credentials) {
+            expect(
+              credentials?.toJson(),
```
Albert221, line 96: Instead of checking `toJson` output (testing it) I'd verify the only the `credentials.tokenEndpoint`

Albert221 (approved):

PR #392, opened 2024-12-10
Add useDisposable hook

n-bernat:
Since `flutter_hooks` doesn't provide a built-in way to dispose resources, and we often use a hook like that in our project (and in packages as well - `leancode_hooks` includes 4 instances of this pattern), I propose adding a new hook that simplifies the disposal of unused resources if an instance of a class supports that.

### DIFF
--- a/packages/leancode_hooks/CHANGELOG.md
+++ b/packages/leancode_hooks/CHANGELOG.md
@@ -1,5 +1,6 @@
-# Unreleased
+# 0.1.0
 
+- Add `useDisposable` hook.
 - Bump `leancode_lint` dev dependency to `12.0.0`.
 - Bump `custom_lint` dev dependency to `0.6.4`.
 
--- a/packages/leancode_hooks/lib/src/use_disposable.dart
+++ b/packages/leancode_hooks/lib/src/use_disposable.dart
@@ -1,0 +1,19 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter_hooks/flutter_hooks.dart';
+
+/// Creates and memoizes an instance created using [builder].
+/// On unmount, executes [dispose] in order to cleanup resources.
+T useDisposable<T>({
+  required ValueGetter<T> builder,
+  required void Function(T) dispose,
+  List<Object?> keys = const <Object>[],
+}) {
+  final value = useMemoized(builder, keys);
+
+  useEffect(
+    () => () => dispose(value),
+    [value],
+  );
+
+  return value;
+}
--- a/packages/leancode_hooks/pubspec.yaml
+++ b/packages/leancode_hooks/pubspec.yaml
@@ -1,13 +1,13 @@
 name: leancode_hooks
-version: 0.0.6
+version: 0.1.0
 homepage: https://github.com/leancodepl/flutter_corelibrary/tree/master/packages/leancode_hooks
 repository: https://github.com/leancodepl/flutter_corelibrary
 description: >-
   Hooks we often use, all gathered in one place.
 
 environment:
-  sdk: '>=3.0.0 <4.0.0'
-  flutter: '>=3.10.0 <4.0.0'
+  sdk: ">=3.0.0 <4.0.0"
+  flutter: ">=3.10.0 <4.0.0"
 
 dependencies:
   bloc: ^8.0.3
--- a/packages/leancode_hooks/README.md
+++ b/packages/leancode_hooks/README.md
@@ -22,6 +22,7 @@ so you won't have to depend on it.
 - [useBlocState](lib/src/use_bloc_state.dart)
 - [useDebounce](lib/src/use_debounce.dart)
 - [useDeclarativeTextEditingController](lib/src/use_decarative_text_editing_controller.dart)
+- [useDisposable](lib/src/use_disposable.dart)
 - [useFocused](lib/src/use_focused.dart)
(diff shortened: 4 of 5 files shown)


packages/leancode_hooks/lib/src/use_disposable.dart
```diff
@@ -0,0 +1,19 @@
...
+T useDisposable<T>({
+  required ValueGetter<T> builder,
+  required void Function(T) dispose,
+  List<Object?> keys = const <Object>[],
+}) {
+  final value = useMemoized(builder, keys);
+
+  useEffect(
+    () => () => dispose(value),
+    [value],
```
PiotrRogulski, line 15: Shouldn't this use the same keys as `useMemoized`, so that it's executed the same number of times?

n-bernat: It wouldn't really change anything, would it? If those keys change, then the instance returned from `useMemoized` would also be different. 

Technically, you can do stuff like extending `ValueNotifier` and override `hashCode` and the equality operator, but then it's your own fault lol

n-bernat: I think I will update it as you mentioned.

PiotrRogulski: I mean, the object can be something different entirely (e.g. a complex data object) that happens to have the same value across recalculations

packages/leancode_hooks/pubspec.yaml
```diff
@@ -1,13 +1,13 @@
...
 homepage: https://github.com/leancodepl/flutter_corelibrary/tree/master/packages/leancode_hooks
 repository: https://github.com/leancodepl/flutter_corelibrary
 description: >-
   Hooks we often use, all gathered in one place.
 
 environment:
-  sdk: '>=3.0.0 <4.0.0'
-  flutter: '>=3.10.0 <4.0.0'
+  sdk: ">=3.0.0 <4.0.0"
+  flutter: ">=3.10.0 <4.0.0"
```
PiotrRogulski, line 10: I think the default style would be single quotes

packages/leancode_hooks/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-# Unreleased
+# 0.1.0
 
+- Add `useDisposable` hook.
 - Bump `leancode_lint` dev dependency to `12.0.0`.
```
PiotrRogulski, line 4: Also on an unrelated note – this is already outdated, but since it hasn't been released yet, maybe we can update it further?

packages/leancode_hooks/pubspec.yaml
n-bernat: Currently, both styles are used in this file, the autoformatter just normalized it.

n-bernat: If that's _the default_, is there any reference for what is considered a default? So, I can at least update my autoformatter

PiotrRogulski: I think Dart uses single quotes by default, if we were to normalize it it should be into single quotes

PiotrRogulski: As for the reference, I just looked at yamls in the repos of Dart and Flutter

n-bernat: I think we may be looking at different repos then? [go_router](https://github.com/flutter/packages/blob/16616ad588d7ad9c00958ec61022e5b5dde5f5dd/packages/go_router/pubspec.yaml#L10), [google_sign_in](https://github.com/flutter/packages/blob/16616ad588d7ad9c00958ec61022e5b5dde5f5dd/packages/google_sign_in/google_sign_in/pubspec.yaml#L10), [image_picker](https://github.com/flutter/packages/blob/16616ad588d7ad9c00958ec61022e5b5dde5f5dd/packages/image_picker/image_picker/pubspec.yaml#L10), [shared_preferences](https://github.com/flutter/packages/blob/16616ad588d7ad9c00958ec61022e5b5dde5f5dd/packages/shared_preferences/shared_preferences/pubspec.yaml#L10), [etc.](https://github.com/search?q=repo%3Aflutter%2Fpackages+%22flutter%3A%22+language%3AYAML&type=code&l=YAML)

PiotrRogulski: Fwiw, if you generate a fresh Flutter project, the pubspec will use single quotes

packages/leancode_hooks/CHANGELOG.md
n-bernat: Could we at least have one release with that version? I want to use stuff from this PR in a project that is stuck with Flutter 3.16 due to the lack of support for Windows 7 in newer versions, and different constraints would require SDK bump 😅

PiotrRogulski: Yeah; I just commented to make a note of this

kotertom:
LGTM. Personally I don't care much about the quote style since we don't have a standardized yaml formatter

kotertom (approved):

packages/leancode_hooks/CHANGELOG.md
kotertom: Can we resolve this thread then? I'll release a new version with the most recent leancode_lint dependency once I'm back from holidays in january, unless someone volunteers to do it earlier :)

PR #393, opened 2024-12-11
[leancode_analytics] Update package configurations

pdenert:
- change getting started in readme
- add gha for publishing


Albert221 (changes requested):
* Add the package to the root's README table
* Add pub and CI shields to the package's README (like in other packages)
* Fix name of the company from `Leancode` that I've never heard of in my life to `LeanCode`

Albert221 (approved):

PR #417, opened 2025-01-17
Refactor caching in Flutter actions

PiotrRogulski:
Also, add a missing build status badge for `leancode_lint` and fix an invalid ignore in `cqrs_test` workflow

### DIFF
--- a/.github/workflows/cqrs-test.yml
+++ b/.github/workflows/cqrs-test.yml
@@ -3,7 +3,7 @@ name: cqrs test
 on:
   push:
     branches: [master]
-    ignore-tags: ['cqrs-v*']
+    tags-ignore: ['cqrs-v*']
     paths:
       - 'packages/cqrs/**'
   pull_request:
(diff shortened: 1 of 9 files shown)


.github/workflows/cqrs-test.yml
```diff
@@ -37,13 +37,6 @@ jobs:
       - name: Dart version
         run: dart --version
 
-      - name: Cache pub dependencies
```
Albert221, line 40: There's no flutter-action so there's no caching. You're removing the only caching step

.github/workflows/login_client-test.yml
```diff
@@ -37,13 +37,6 @@ jobs:
       - name: Dart version
         run: dart --version
 
-      - name: Cache pub dependencies
```
Albert221, line 40: There's no flutter-action so there's no caching. You're removing the only caching step

Albert221 (approved):

PR #420, opened 2025-01-29
[LMG-109] Publish force update

michalina-majewska:
I've sorted packages in README alphabetically, I hope that's okay.  I suggest ignoring this commit when reviewing.


Albert221 (commented):
* The minimum (and only) version we run tests for the package is 3.24, but pubspec allows 3.10. Please bump or debump one of these.

Albert221 (approved):

PR #434, opened 2025-02-16
Use positional arguments instead of manual string interpolation

PiotrRogulski:
This PR is purely technical and does not introduce any behavior changes.

Refactors lint codes to use built-in positional parameters (like `{0}`) in lint messages.

### DIFF
--- a/packages/leancode_lint/CHANGELOG.md
+++ b/packages/leancode_lint/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+- chore: Refactor lint codes to use built-in positional parameters.
+
 # 15.0.0
 
 - Enable the following lints:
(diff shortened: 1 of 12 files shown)


shilangyu:
Personally, I am not convinced to the benefits. This makes it much more susceptible to refactors. It is very easy to desync `arguments:` and `{n}` placeholders.

Yes, it is nice that it is built in. But Dart APIs are known to be designed in a pragmatic (???) rather than type safe way. We can build a small type safe wrapper?

PiotrRogulski:
> Personally, I am not convinced to the benefits. This makes it much more susceptible to refactors. It is very easy to desync `arguments:` and `{n}` placeholders.
> 
> Yes, it is nice that it is built in. But Dart APIs are known to be designed in a pragmatic (???) rather than type safe way. We can build a small type safe wrapper?

It is indeed unfortunate that the analyzer only accepts positional arguments. The main benefit is that we define the lints the way they're supposed to be defined. In particular, after this refactor, we create only one LintCode with the template string – this way, it's actually easier to keep the messages consistent, in my opinion

shilangyu (commented):
Oh well, I don't see any better way and I like doing it the intended way

Artur-Wisniewski (approved):
LGTM!

PR #452, opened 2025-03-10
Make CommandResponse public

gogolon:
`CommandResponse` (formerly `CommandResult`) is used in contracts, so it has to be accessible from outside of this package ([docs](https://github.com/leancodepl/contractsgenerator/blob/f3f2d112f4368f9616e4e7a9e5b90950db25a35b/docs/types.md)).

### DIFF
--- a/packages/cqrs/CHANGELOG.md
+++ b/packages/cqrs/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 10.0.2
 
-- Make CommandResponse publicly accessible.
+- Make CommandResponse public.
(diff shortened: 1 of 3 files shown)


packages/cqrs/CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-# Unreleased
+# 10.0.2
 
+- Make CommandResponse publicly accessible.
```
Albert221, line 3: public

Albert221 (approved):

PR #458, opened 2025-05-09
New bloc-related lints

PiotrRogulski:
This PR introduces six new lints enforcing our new architecture guidelines.

(Also, as a cleanup, all created single-type checkers are moved to one location to avoid duplication.)

### DIFF
--- a/packages/leancode_lint/lib/lints/bloc_related_classes_equatable.dart
+++ b/packages/leancode_lint/lib/lints/bloc_related_classes_equatable.dart
@@ -16,7 +16,7 @@ class BlocRelatedClassesEquatable extends DartLintRule {
       );
 
   @override
-  List<Fix> getFixes() => [AddMixin()];
+  List<Fix> getFixes() => [_AddMixin()];
 
   @override
   void run(
@@ -50,7 +50,7 @@ class BlocRelatedClassesEquatable extends DartLintRule {
   }
 }
 
-class AddMixin extends DartFix {
+class _AddMixin extends DartFix {
   @override
   void run(
     CustomLintResolver resolver,
--- a/packages/leancode_lint/lib/lints/use_equatable_mixin.dart
+++ b/packages/leancode_lint/lib/lints/use_equatable_mixin.dart
@@ -16,7 +16,7 @@ class UseEquatableMixin extends DartLintRule {
       );
 
   @override
-  List<Fix> getFixes() => [ConvertToMixin()];
+  List<Fix> getFixes() => [_ConvertToMixin()];
 
   @override
   void run(
@@ -53,7 +53,7 @@ class UseEquatableMixin extends DartLintRule {
   }
 }
 
-class ConvertToMixin extends DartFix {
+class _ConvertToMixin extends DartFix {
   @override
   void run(
     CustomLintResolver resolver,


packages/leancode_lint/lib/lints/bloc_class_modifiers.dart
```diff
@@ -0,0 +1,60 @@
...
+      );
+
+  @override
+  void run(
+    CustomLintResolver resolver,
+    ErrorReporter reporter,
+    CustomLintContext context,
+  ) {
+    context.registry.addBloc((node, data) {
+      void checkHierarchy(InterfaceElement? element) {
```
Artur-Wisniewski, line 24: InterfaceElement is deprecated. Check if you can use another element class.

packages/leancode_lint/lib/helpers.dart
```diff
@@ -218,6 +217,92 @@ extension LintRuleNodeRegistryExtensions on LintRuleNodeRegistry {
       listener(buildMethod.body, diagnosticNode);
     });
   }
+
+  void addBloc(void Function(ClassDeclaration node, _BlocData data) listener) {
```
Artur-Wisniewski, line 221: We could add some documentation here. AddBloc is, if I am not wrong, limited to one file. So _BlocData data wouldn't be correct fully. So there is an assumption that the whole bloc with events, states, is in one file. It would be worth mentioning this.

packages/leancode_lint/lib/lints/bloc_const_constructors.dart
```diff
@@ -0,0 +1,46 @@
+import 'package:analyzer/error/error.dart' hide LintCode;
+import 'package:analyzer/error/listener.dart';
+import 'package:custom_lint_builder/custom_lint_builder.dart';
+import 'package:leancode_lint/helpers.dart';
+
+class BlocConstConstructors extends DartLintRule {
```
Artur-Wisniewski, line 6: What do you think about adding DartFix for this lint? Lint that will create a const constructor. 
It could be problematic if Bloc has variables, but I think we have to only handle case where there are no parameters in Bloc's body. And in this case, we only have to create an empty const constructor

packages/leancode_lint/README.md
```diff
@@ -413,6 +413,203 @@ class MyWidget extends StatelessWidget {
...
+final class MyStateLoading extends MyState {}
+```
+
+#### Configuration
+
+None.
+
+### `bloc_const_constructors`
+
+**DO** define unnamed const constructors for bloc state, event, and presentation event classes.
```
Artur-Wisniewski, line 446: Ditto, it would be nice to highlight that this lint works only if bloc is in the same file as states, events etc

packages/leancode_lint/test/lints_test_app/lib/bloc_const_constructors_test.dart
```diff
@@ -0,0 +1,66 @@
...
+  const Test3Event();
+}
+
+class Test3State {
+  const Test3State();
+}
+
+class Test3PresentationEvent {
+  const Test3PresentationEvent();
+}
```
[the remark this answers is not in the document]
PiotrRogulski: I don't see a straightforward way to check if a constructor can be made const, so if anything, I would indeed do it separately

packages/leancode_lint/test/lints_test_app/lib/bloc_related_classes_equatable_test.dart
```diff
@@ -0,0 +1,84 @@
...
+// expect_lint: bloc_related_classes_equatable
+Test1State {}
+
+final class
+// expect_lint: bloc_related_classes_equatable
+Test1PresentationEvent {}
+
+//////////////////////////////////////////////////////////////////////////
+
+// Bloc-related classes that `extend Equatable` are not flagged
```
Artur-Wisniewski, line 28: I think it would be better if we flag them also. Using Equatable is more problematic than using a mixin version. Maybe we could do such a flag with `extends` equatable as an info severity instead of a warning? I wonder if there is any case where `extends Equatable` is a considerably more suitable solution.

packages/leancode_lint/test/lints_test_app/lib/bloc_related_classes_equatable_test.dart
```diff
@@ -0,0 +1,84 @@
...
+  @override
+  List<Object?> get props => [];
+}
+
+final class Test2State extends Equatable {
+  @override
+  List<Object?> get props => [];
+}
+
+final class Test2PresentationEvent extends Equatable {
```
Artur-Wisniewski, line 44: I am not fully convinced that we should use Equatable if there aren't variables in the state

packages/leancode_lint/lib/lints/bloc_related_classes_equatable.dart
```diff
@@ -0,0 +1,73 @@
...
+      }
+
+      check(data.stateElement);
+      check(data.eventElement);
+      check(data.presentationEventElement);
+    });
+  }
+}
+
+class AddMixin extends DartFix {
```
Artur-Wisniewski, line 53:
```suggestion
class _AddMixin extends DartFix {
```

packages/leancode_lint/test/lints_test_app/lib/bloc_related_class_naming_test.dart
```diff
@@ -0,0 +1,100 @@
...
+  SamePackageBloc() : super(ClassFromSamePackage());
+}
+
+//////////////////////////////////////////////////////////////////////////
+
+// No lint when class is from another package
+class AnotherPackageBloc extends Bloc<int, int>
+    with BlocPresentationMixin<int, int> {
+  AnotherPackageBloc() : super(0);
+}
```
Artur-Wisniewski, line 100: Please add a case where we have more than one StateClass eg:
```
class Correct1Cubit extends Cubit<Correct1State>
    with BlocPresentationMixin<Correct1State, Correct1Event> {
  Correct1Cubit() : super(Correct1State());
}

selaed class Correct1State {}

final class Correct1Initial {}
final class Correct1Loading {}
final class Correct1Loaded {}
```

Artur-Wisniewski: Additionally I think: 
```
final class Correct1Initial {}
final class Correct1Loading {}
final class Correct1Loaded {}
```
These cases should be legal.

packages/leancode_lint/test/lints_test_app/lib/bloc_related_classes_equatable_test.dart
Artur-Wisniewski: Yes, I've seen it later, unfortunately. Good work :)

packages/leancode_lint/test/lints_test_app/lib/bloc_related_class_naming_test.dart
Artur-Wisniewski: Ok makes sense that it follows leancode standards but I could be misleading for external people because in bloc documentation examples uses this format:
```
selaed class ExampleState {}

final class ExampleInitial {}
final class ExampleLoading {}
final class ExampleLoaded {}
```
Maybe we could add some configuration to allow this?

packages/leancode_lint/lib/helpers.dart
```diff
@@ -218,6 +218,90 @@ extension LintRuleNodeRegistryExtensions on LintRuleNodeRegistry {
...
+
+  final stateType = blocElement.allSupertypes
+      .firstWhere(TypeCheckers.blocBase.isExactlyType)
+      .typeArguments
+      .singleOrNull;
+  if (stateType == null) {
+    return null;
+  }
+
+  final stateElement = stateType.element3;
```
Artur-Wisniewski, line 257: image
What happened here ;_;
Why `Element2? get element3` 😆

packages/leancode_lint/lib/lints/use_equatable_mixin.dart
```diff
@@ -0,0 +1,74 @@
+import 'package:analyzer/error/error.dart' hide LintCode;
+import 'package:analyzer/error/listener.dart';
+import 'package:analyzer/source/source_range.dart';
+import 'package:custom_lint_builder/custom_lint_builder.dart';
+import 'package:leancode_lint/common_type_checkers.dart';
+
+class UseEquatableMixin extends DartLintRule {
```
Artur-Wisniewski, line 7: Rename class to match lint name

Artur-Wisniewski (commented):
Wow great lints! I've added some comments, but I think you've done a decent job here. Please address comments above :)

PR #471, opened 2025-08-02
feat: Adding new use_padding lint

Artur-Wisniewski:


packages/leancode_lint/test/lints_test_app/lib/use_padding_test.dart
```diff
@@ -0,0 +1,39 @@
+// Ignored for test purposes
+// ignore_for_file: avoid_redundant_argument_values
+
+import 'package:flutter/material.dart';
+
+class UsePaddingTest {
+  void test() {
+    // expect_lint: use_padding
+    Container(margin: const EdgeInsets.all(10), child: const SizedBox());
```
PiotrRogulski, line 9: I just realized something: please reformat these tests so that the expect is above the line with the target node only (so here, ensure that the `Container` text is on its own line). The same applies to the other tests you've already written

PiotrRogulski: Yes — see `avoid_single_child_in_multi_child_widget_test`

packages/leancode_lint/test/lints_test_app/lib/use_padding_test.dart
```diff
@@ -0,0 +1,60 @@
+// Ignored for test purposes
+// ignore_for_file: avoid_redundant_argument_values
+
+import 'package:flutter/material.dart';
+
+// dart format off
```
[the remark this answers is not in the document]
Artur-Wisniewski: Docs saying:
The comments must be exactly // dart format off and // dart format on. A file may have multiple regions, but they can't overlap or nest. `If you want to opt the rest of a file out, you can use a single // dart format off comment without a closing // dart format on at the end`.

But I've added `// dart format on` for cleaner code.

packages/leancode_lint/lib/lints/use_padding.dart
```diff
@@ -0,0 +1,75 @@
...
+import 'package:leancode_lint/helpers.dart';
+import 'package:leancode_lint/utils.dart';
+
+class UsePadding extends DartLintRule {
+  const UsePadding()
+    : super(
+        code: const LintCode(
+          name: ruleName,
+          problemMessage:
+              'Use Padding widget instead of the Container widget with only the padding parameter',
```
PiotrRogulski, line 15:
```suggestion
              'Use Padding widget instead of the Container widget with only the margin parameter',
```

PR #494, opened 2025-11-25
Migrate lints to analysis server plugin

PiotrRogulski:
In separate PRs:
- ~re-write tests~
- move implementation to `src`

What's not yet available:
- ~`use_design_system_item`: separate lint rules per config entry~ managed to get it back with the new config approach
- `prefer_center_over_align`: our current implementation requires an asynchronous resolution of another library, which isn't (yet?) possible

(cc @shilangyu, I can't add you as a reviewer 🙈)

### DIFF
--- a/packages/leancode_lint/test/mock_libraries.dart
+++ b/packages/leancode_lint/test/mock_libraries.dart
@@ -1,96 +1,16 @@
 import 'package:analyzer_testing/analysis_rule/analysis_rule.dart';
 
-enum MockLibrary {
-  bloc('bloc', {'lib/bloc.dart': _blocMock}),
-  flutterBloc('flutter_bloc', {'lib/flutter_bloc.dart': _flutterBlocMock}),
-  sliverTools('sliver_tools', {'lib/sliver_tools.dart': _sliverToolsMock}),
-  flutterHooks('flutter_hooks', {'lib/flutter_hooks.dart': _flutterHooksMock}),
-  hooksRiverpod('hooks_riverpod', {
-    'lib/hooks_riverpod.dart': _hooksRiverpodMock,
-  });
+part 'mock_libraries/bloc.dart';
+part 'mock_libraries/flutter.dart';
+part 'mock_libraries/flutter_bloc.dart';
+part 'mock_libraries/flutter_hooks.dart';
+part 'mock_libraries/hooks_riverpod.dart';
+part 'mock_libraries/sliver_tools.dart';
 
-  const MockLibrary(this.name, this.files);
-
-  final String name;
-  final Map<String, String> files;
-}
-
-extension AddMockLibraryX on AnalysisRuleTest {
-  void addMocks(Iterable<MockLibrary> mocks) {
-    for (final mock in mocks) {
-      final package = newPackage(mock.name);
-      for (final MapEntry(key: path, value: content) in mock.files.entries) {
-        package.addFile(path, content);
-      }
-    }
+extension on AnalysisRuleTest {
+  void appendFile(String path, String content) {
+    final file = getFile(path);
+    final oldContent = file.readAsStringSync();
+    file.writeAsStringSync('$oldContent\n$content');
   }
 }
-
-const _blocMock = '''
-abstract class Cubit<State> extends BlocBase<State> {
-  Cubit(State initialState) : super(initialState);
-}
-''';
-
-const _flutterBlocMock = '''
-export 'package:bloc/bloc.dart';
-''';
-
-const _sliverToolsMock = '''
-import 'package:flutter/material.dart';
-
-class MultiSliver extends StatelessWidget {
-  MultiSliver({required List<Widget> children});
-}
-''';
-
-const _flutterHooksMock = '''
-import 'package:flutter/material.dart';
-
-void useAutomaticKeepAlive() {}
-
-ValueNotifier<T> useState<T>(T initialData) {}
-
-TextEditingController useTextEditingController() {}
-
-PageController usePageController() {}
-
-class HookWidget extends StatelessWidget {
-  const HookWidget({super.key});
-}
-
-class HookBuilder extends HookWidget {
-  const HookBuilder({required Widget Function(BuildContext context) builder});
-}
-
-BuildContext useContext() {}
-
-T useMemoized<T>(
-  T Function() valueBuilder, [
-  List<Object?> keys = const <Object>[],
-]) {}
-''';
-
-const _hooksRiverpodMock = '''
-import 'package:flutter/material.dart';
-
-class WidgetRef {}
-
-abstract class ConsumerWidget extends StatefulWidget {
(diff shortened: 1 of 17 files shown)


PiotrRogulski:
> What is your opinion on how the tests look like now?

Not particularly fond of them, honestly 😅

> they seem not maintainable and not made for humans to edit

That's not a huge issue; the testing tooling suggests appropriate locations:

```
Found but did not expect:
  use_design_system_item.use_design_system_item [57, 8, Scaffold is forbidden within this design system., "Use the alternative defined in the design system: LftScaffold."]
  use_design_system_item.use_design_system_item [106, 4, Text is forbidden within this design system., "Use the alternative defined in the design system: LftText."]

To accept the current state, expect:
  lint(57, 8),
  lint(106, 4),
```

> Either way, I guess this is not mergable until we figure out how to support configs (which are crucial)

I mean, the config, even if not done "officially", already works

packages/leancode_lint/lib/assists/convert_iterable_map_to_collection_for.dart
shilangyu: Does it work for you? For me not

packages/leancode_lint/lib/assists/convert_positional_to_named_formal.dart
shilangyu: This does not work either for me

packages/leancode_lint/lib/assists/convert_record_into_nominal_type.dart
shilangyu: This does work

packages/leancode_lint/test/test_cases/add_cubit_suffix_for_cubits_test.dart
```diff
@@ -21,31 +22,28 @@ class AddCubitSuffixForYourCubitsTest extends AnalysisRuleTest {
...
+class /*[1*/NotProperlyNamedWithDifferentSuperClass/*1]*/ extends ProperlyNamedCubit {
   NotProperlyNamedWithDifferentSuperClass();
 }
 
 class ProperlyNamedWithDifferentSuperClassCubit extends ProperlyNamedCubit {
   ProperlyNamedWithDifferentSuperClassCubit();
 }
 
-class ClassExtendingOtherCubitClassWithoutSuffix extends NotProperlyNamed {
+class /*[2*/ClassExtendingOtherCubitClassWithoutSuffix/*2]*/ extends NotProperlyNamed {
```
shilangyu: What is the difference between this and `[!!]`? `[!!]` works if there is only one diagnostic?

packages/leancode_lint/lib/lints/start_comments_with_space.dart
```diff
@@ -80,22 +87,24 @@ enum _CommentType {
   final String pluralName;
 }
 
-class _AddStartingSpaceToComment extends DartFix {
+class AddStartingSpaceToComment extends ResolvedCorrectionProducer {
```
shilangyu: Do you know if quick fixes are easily testable now?

packages/leancode_lint/lib/config.dart
```diff
@@ -0,0 +1,57 @@
+import 'package:analyzer/analysis_rule/rule_context.dart';
+// TODO: use a canonical way to access analysis options (https://github.com/dart-lang/sdk/issues/61755)
+// ignore: implementation_imports
+import 'package:analyzer/src/workspace/pub.dart';
+import 'package:yaml/yaml.dart';
+
+class LeancodeLintConfig {
+  LeancodeLintConfig._(this.configMap);
+
+  factory LeancodeLintConfig.fromRuleContext(RuleContext context) {
```
shilangyu: This unfortunately reads only analysis_options from the root of the workspace. It is an important feature (and used at LeanCode) that configs are scoped to some directory.

`package:analyzer/src/analysis_options` handles this. See the `AnalysisOptionsProvider` class. It also handles merging options from upper configs and resolves `include:` directives.

packages/leancode_lint/test/test_cases/add_cubit_suffix_for_cubits_test.dart
PiotrRogulski: Yes, `[!` is an alias for `/*[0*/`. There's also `^`, which is an alias for the position marker `/*0*/`.

packages/leancode_lint/lib/assists/convert_iterable_map_to_collection_for.dart
PiotrRogulski: It turns out that the change builder was simply not awaited 🤡

Screenshot 2025-12-07 at 22 34 06

packages/leancode_lint/test/mock_libraries.dart
```diff
@@ -0,0 +1,380 @@
...
+        from_package: flutter
+      - instead_of: RichText
+        from_package: flutter
+    LftScaffold:
+      - instead_of: Scaffold
+        from_package: flutter
+    LftPlaceholder:
+      - instead_of: Placeholder
+        from_package: flutter
+''');
```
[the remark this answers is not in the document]
PiotrRogulski: Hmm, where do you suggest I put it? This is just a sample config to test config-aware lints

packages/leancode_lint/lib/config.dart
```diff
@@ -0,0 +1,60 @@
...
+    final package = context.package;
+    if (package is! PubPackage) {
+      return _empty;
+    }
+
+    final provider = AnalysisOptionsProvider(
+      .new([package.workspace.packageUriResolver]),
+    );
+
+    final analysisOptionsFile = provider.getOptionsFile(
```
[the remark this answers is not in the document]
shilangyu: I spent some time debugging this. It really seems like a bug. I tried to look at how the SDK uses this API but found nothing jarring. I hope to find more time this week (but it is very busy for me) to resolve this finally.

packages/leancode_lint/lib/lints/prefer_center_over_align.dart
```diff
@@ -1,141 +1,143 @@
-import 'package:analyzer/dart/analysis/results.dart';
-import 'package:analyzer/dart/ast/ast.dart';
-import 'package:analyzer/dart/constant/value.dart';
-import 'package:analyzer/diagnostic/diagnostic.dart';
-import 'package:analyzer/error/listener.dart';
-import 'package:analyzer_plugin/utilities/range_factory.dart';
-import 'package:custom_lint_builder/custom_lint_builder.dart';
-import 'package:leancode_lint/helpers.dart';
+// TODO: decide what to do with this lint in light of https://github.com/dart-lang/sdk/issues/61517
```
shilangyu: What I gathered from the SDK team responses is that there is no good way to resolve this. I think we need to give up on computeConstantValue approach and just hardcode some cases. Or, I can try the proposed approach of crawling from the `Align` widget to the `Alignment.center` constant. I wonder how expensive this is.

Issue #465, opened 2025-06-13
[leancode_lint] Allow custom parameter names in `catch_parameter_names`  lint


nathnaeld:
Currently, the `catch_parameter_names` lint rule restricts parameter names to `err` and `st`. To increase flexibility and accommodate various coding standards, it would be beneficial to support custom parameter names in the lint configuration.

mchudy:
Thanks for raising this issue — it’s indeed a good suggestion to provide some configuration options to this lint. We’re open to the idea and will consider adding support for custom parameter names in the config. That said, we like to use one standard for all of our projects and we're not planning to change it right now, so if you need it sooner, feel free to open a PR — we’d really appreciate it!

PR #496, opened 2025-12-28
Add configurable names for `catch_parameter_names` rule

natty-pluz:
**Related issue:** [#465 Allow custom parameter names in `catch_parameter_names` rule](https://github.com/leancodepl/flutter_corelibrary/issues/465)

## Summary

Add support for customizing the parameter names enforced by the `catch_parameter_names` lint. Defaults remain `err` (exception) and `st` (stack trace); projects can override them via YAML.

## Why

The rule previously enforced only `err`/`st`, which is too strict for some codebases. Making names configurable increases flexibility while preserving the default style.

## Config example

```yaml
- catch_parameter_names:
  exception: error
  stack_trace: stackTrace
```

## What changed

* Add `CatchParameterNamesConfig` and wire it into the lint rule.
* Use configured names when present; fall back to `err` / `st`.
* Keep accepting `_` as a valid (ignored) parameter name.
* Short doc updated to mention the config.

## Compatibility

Backwards compatible: existing projects that do not add config keep current behavior.

### DIFF
--- a/packages/leancode_lint/lib/lints/catch_parameter_names.dart
+++ b/packages/leancode_lint/lib/lints/catch_parameter_names.dart
@@ -98,7 +98,6 @@ class CatchParameterNames extends DartLintRule {
     final actualName = node.name.lexeme;
     final preferred = config.preferredName(param);
 
-    // accept underscore or the preferred name
     if (actualName == '_' || actualName == preferred) {
       return;
     }


mchudy (approved):
LGTM with a small nit. Thanks a lot for your contribution, really appreciated!

PR #530, opened 2026-05-14
Add missing_equatable_props lint rule

alszczep:


packages/leancode_lint/lib/src/lints/missing_equatable_props.dart
```diff
@@ -0,0 +1,276 @@
...
+    RuleVisitorRegistry registry,
+    RuleContext context,
+  ) {
+    registry.addClassDeclaration(this, _Visitor(this, context));
+  }
+}
+
+const _equatableTypeChecker = TypeChecker.any([
+  TypeChecker.fromName('Equatable', packageName: 'equatable'),
+  TypeChecker.fromName('EquatableMixin', packageName: 'equatable'),
```
PiotrRogulski:
```suggestion
  .fromName('Equatable', packageName: 'equatable'),
  .fromName('EquatableMixin', packageName: 'equatable'),
```

packages/leancode_lint/lib/src/lints/missing_equatable_props.dart
```diff
@@ -0,0 +1,276 @@
...
+    final missingFields = _findMissingFieldNames(members, contents.names);
+
+    if (!needsSuper && missingFields.isEmpty) {
+      return;
+    }
+
+    await builder.addDartFileEdit(file, (builder) {
+      if (listLiteral.elements.isEmpty) {
+        final text = [
+          if (needsSuper) '...super.props',
```
PiotrRogulski: `super.props` can be inserted without spreading

packages/leancode_lint/lib/src/lints/missing_equatable_props.dart
```diff
@@ -0,0 +1,276 @@
...
+      if (needsSuper) {
+        builder.addSimpleInsertion(
+          listLiteral.elements.first.offset,
+          '...super.props, ',
+        );
+      }
+      if (missingFields.isNotEmpty) {
+        builder.addSimpleInsertion(
+          listLiteral.elements.last.end,
+          ', ${missingFields.join(', ')}',
```
PiotrRogulski, line 141: If the list is multi-line, there'll be a trailing comma – does this work correctly?

packages/leancode_lint/lib/src/lints/missing_equatable_props.dart
```diff
@@ -0,0 +1,276 @@
...
+        names.add(name);
+      case PropertyAccess(target: ThisExpression(), :final propertyName):
+        names.add(propertyName.name);
+      case SpreadElement(
+        expression: PropertyAccess(
+          target: SuperExpression(),
+          propertyName: SimpleIdentifier(name: 'props'),
+        ),
+      ):
+        hasSuperPropsSpread = true;
```
PiotrRogulski: Simplarly to my other comment: `super.props` itself without the stread should be allowed as `hasSuperPropsSpread` as well

packages/leancode_lint/lib/src/lints/missing_equatable_props.dart
```diff
@@ -0,0 +1,276 @@
...
+/// `props` list literal. When the class extends another Equatable-like class,
+/// `...super.props` must also be present so that inherited fields participate
+/// in equality.
+class MissingEquatableProps extends AnalysisRule {
+  MissingEquatableProps()
+    : super(name: code.lowerCaseName, description: code.problemMessage);
+
+  static const code = LintCode(
+    'missing_equatable_props',
+    "The following are missing from Equatable's props: {0}.",
```
PiotrRogulski:
```suggestion
    "Class is missing the following Equatable props: {0}.",
```

packages/leancode_lint/lib/src/lints/missing_equatable_props.dart
```diff
@@ -0,0 +1,276 @@
...
+      return;
+    }
+
+    final listLiteral = _getPropsListLiteral(propsGetter);
+    if (listLiteral == null) {
+      return;
+    }
+
+    final contents = _collectExistingProps(listLiteral);
+    if (contents.hasUnrecognizedContent) {
```
PiotrRogulski: I don't think this is needed, if `hasUnrecognizedContent` is true, the lint won't be reported at all (so the fix won't run)

packages/leancode_lint/lib/src/lints/missing_equatable_props.dart
```diff
@@ -0,0 +1,276 @@
...
+    final members = _getMembers(classDecl);
+    final propsGetter = _findPropsGetter(members);
+    if (propsGetter == null) {
+      return;
+    }
+
+    final listLiteral = _getPropsListLiteral(propsGetter);
+    if (listLiteral == null) {
+      return;
+    }
```
PiotrRogulski: Ditto, let's not re-check what we already checked in the lint itself

packages/leancode_lint/lib/src/lints/missing_equatable_props.dart
alszczep: Looks good to me, if the trailing comma was there then it will be there after the fix, but won't be added if it wasn't. Also I'd assume that this doesn't have to always be perfect, bcs after the fix quite often you'd format the file.

packages/leancode_lint/lib/plugin.dart
```diff
@@ -82,6 +83,11 @@ final class LeanCodeLintPlugin extends Plugin {
         PreferEquatableMixin.code,
         ConvertToEquatableMixin.new,
       )
+      ..registerWarningRule(MissingEquatableProps())
```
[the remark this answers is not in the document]
PiotrRogulski: We could indeed add a changelog entry for the next release

packages/leancode_lint/lib/src/lints/missing_equatable_props.dart
```diff
@@ -0,0 +1,259 @@
...
+    'Add all missing entries to props',
+  );
+
+  @override
+  CorrectionApplicability get applicability =>
+      CorrectionApplicability.automatically;
+
+  @override
+  Future<void> compute(ChangeBuilder builder) async {
+    final classDecl = node.thisOrAncestorOfType<ClassDeclaration>()!;
```
PiotrRogulski: Don't use abbreviations (like Decl)

packages/leancode_lint/lib/src/lints/missing_equatable_props.dart
```diff
@@ -0,0 +1,259 @@
...
+    if (needsSuper) 'super.props',
+    ..._findMissingFieldNames(members, contents.names),
+  ];
+}
+
+List<String> _findMissingFieldNames(
+  List<ClassMember> members,
+  Set<String> existingNames,
+) => [
+  for (final fieldDecl in members.whereType<FieldDeclaration>())
```
PiotrRogulski: Ditto

packages/leancode_lint/lib/src/lints/missing_equatable_props.dart
```diff
@@ -0,0 +1,259 @@
...
+bool _shouldHaveSuperProps(InterfaceElement element) {
+  final supertype = element.supertype;
+  if (supertype == null) {
+    return false;
+  }
+  final supertypeElement = supertype.element;
+  if (_equatableTypeChecker.isExactly(supertypeElement)) {
+    return false;
+  }
+  return _equatableTypeChecker.isAssignableFrom(supertypeElement);
```
PiotrRogulski: Hmm, what if the superclass extends Equatable, but is itself abstract, and thus doesn't define concrete props? Referencing them isn't valid

packages/leancode_lint/lib/plugin.dart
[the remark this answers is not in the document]
PiotrRogulski: @alszczep bump

packages/leancode_lint/test/test_cases/missing_equatable_props_test.dart
```diff
@@ -0,0 +1,348 @@
...
+}
+
+class Sub extends AbstractBase {
+  const Sub(this.a, this.b);
+
+  final int a;
+  final int b;
+
+  @override
+  List<Object?> get props => /*[0*/[a]/*0]*/;
```
PiotrRogulski, line 284: Please add a test: there in an abstract equatable base class, and the derived class has all the props => no diagnostic

PiotrRogulski (approved):

PR #540, opened 2026-06-29
[leancode_debug_page] Upgrade `share_plus`, fix sharing on Darwin systems

PiotrRogulski:
Because of `share_plus`, SDK constraints had to be upgraded:

```
Because leancode_debug_page depends on share_plus >=13.0.0 which requires SDK version >=3.10.0 <4.0.0, version solving failed.
```


mateusz-pietras (approved):
LGTM

PR #545, opened 2026-07-14
Add `avoid_build_context_in_blocs` lint

cupofme:
Adds a warning-level lint that flags `BuildContext` crossing into a Bloc/Cubit — either passed in (via a method like `add`, or a constructor) or declared inside one (parameter or field).

### DIFF
--- a/packages/leancode_lint/lib/src/lints/avoid_build_context_in_blocs.dart
+++ b/packages/leancode_lint/lib/src/lints/avoid_build_context_in_blocs.dart
@@ -1,0 +1,231 @@
+import 'package:analyzer/analysis_rule/analysis_rule.dart';
+import 'package:analyzer/analysis_rule/rule_context.dart';
+import 'package:analyzer/analysis_rule/rule_visitor_registry.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:analyzer/error/error.dart';
+import 'package:leancode_lint/src/bloc_utils.dart';
+import 'package:leancode_lint/src/type_checker.dart';
+
+/// Warns when a `BuildContext` crosses into a Bloc/Cubit.
+///
+/// A `BuildContext` couples business logic to the widget tree, which risks
+/// stale contexts and wrong `InheritedWidget` reads and makes the logic hard to
+/// test. This rule flags a `BuildContext` on both sides of the boundary:
+/// passing one into a Bloc/Cubit (via a method, e.g. `bloc.add(...)`, or a
+/// constructor) and declaring one inside a Bloc/Cubit (as a parameter or field).
+class AvoidBuildContextInBlocs extends AnalysisRule {
+  AvoidBuildContextInBlocs()
+    : super(name: code.lowerCaseName, description: code.problemMessage);
+
+  static const code = LintCode(
+    'avoid_build_context_in_blocs',
+    // The specific clause is supplied per report site via `arguments`.
+    '{0}',
+    correctionMessage:
+        "Business logic shouldn't depend on the widget tree. Pass only the data the Bloc/Cubit needs.",
+    severity: .WARNING,
+  );
+
+  @override
+  LintCode get diagnosticCode => code;
+
+  @override
+  void registerNodeProcessors(
+    RuleVisitorRegistry registry,
+    RuleContext context,
+  ) {
+    final visitor = _Visitor(this);
+    registry
+      ..addMethodInvocation(this, visitor)
+      ..addInstanceCreationExpression(this, visitor)
+      ..addClassDeclaration(this, visitor);
+  }
+}
+
+const _passingMessage = "Avoid passing 'BuildContext' to blocs.";
+
+String _parameterMessage(BlocType type) => switch (type) {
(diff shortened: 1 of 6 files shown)


packages/leancode_lint/lib/src/lints/avoid_build_context_in_blocs.dart
```diff
@@ -0,0 +1,234 @@
...
+        name: code.lowerCaseName,
+        description: "Avoid letting a 'BuildContext' cross into a Bloc/Cubit.",
+      );
+
+  static const code = LintCode(
+    'avoid_build_context_in_blocs',
+    // The specific clause is supplied per report site via `arguments`.
+    '{0}',
+    correctionMessage:
+        "Business logic shouldn't depend on the widget tree. Pass only the data the Bloc/Cubit needs.",
```
PiotrRogulski: The correction message should be exactly that: What to do to make the code compliant with the lint. This current string is more of a description (at least the first part)

packages/leancode_lint/lib/src/lints/avoid_build_context_in_blocs.dart
```diff
@@ -0,0 +1,234 @@
...
+  AvoidBuildContextInBlocs()
+    : super(
+        name: code.lowerCaseName,
+        description: "Avoid letting a 'BuildContext' cross into a Bloc/Cubit.",
+      );
+
+  static const code = LintCode(
+    'avoid_build_context_in_blocs',
+    // The specific clause is supplied per report site via `arguments`.
+    '{0}',
```
PiotrRogulski: Let's not supply the ***entire*** message as the arg. Make it a template here and substitute as little as possible. Also, reuse it in the super constructor

packages/leancode_lint/lib/src/lints/avoid_build_context_in_blocs.dart
cupofme: Updated to `Remove the 'BuildContext' and directly pass only the data the Bloc/Cubit needs instead.`

packages/leancode_lint/lib/src/lints/avoid_build_context_in_blocs.dart
cupofme: Simplified to just `Avoid using 'BuildContext' in a Bloc/Cubit.`

PiotrRogulski: Hmm, I don't like that too much 🫥 Let's customize the message at least a little bit, to make it more relevant to the reported location:

- `Bloc/Cubit` – we can use the detected type here
- `Avoid using` – "using" is very vague; let's say what kind of usage we detected

packages/leancode_lint/lib/src/lints/avoid_build_context_in_blocs.dart
```diff
@@ -0,0 +1,217 @@
...
+          break;
+      }
+    }
+  }
+
+  void _checkParameters(FormalParameterList parameters) {
+    for (final parameter in parameters.parameters) {
+      final name = parameter.name;
+      final type = parameter.declaredFragment?.element.type;
+      if (name != null && type != null && _isBuildContext(type)) {
```
PiotrRogulski: Use pattern matching on `parameter` to do null checks and bind to variables

packages/leancode_lint/lib/src/lints/avoid_build_context_in_blocs.dart
```diff
@@ -0,0 +1,217 @@
...
+    Expression? expression,
+    Set<Element> visited,
+    int depth,
+  ) {
+    if (expression == null || depth > 20) {
+      return false;
+    }
+
+    final type = expression.staticType;
+    if (type != null && _isBuildContext(type)) {
```
PiotrRogulski:
```suggestion
    if (expression.staticType case final type? when _isBuildContext(type)) {
```

packages/leancode_lint/lib/src/lints/avoid_build_context_in_blocs.dart
```diff
@@ -0,0 +1,217 @@
...
+          if (_carriesContext(
+            argument.argumentExpression,
+            visited,
+            depth + 1,
+          )) {
+            return true;
+          }
+        }
+      case SimpleIdentifier(:final Element element?)
+          when element is LocalVariableElement && visited.add(element):
```
PiotrRogulski:
```suggestion
      case SimpleIdentifier(:final LocalVariableElement element?)
          when visited.add(element):
```

packages/leancode_lint/lib/src/lints/avoid_build_context_in_blocs.dart
cupofme: Made it more context-aware

packages/leancode_lint/lib/src/lints/avoid_build_context_in_blocs.dart
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  bool _isBuildContext(DartType type) =>
+      _buildContextChecker.isExactlyType(type);
+}
+
+String _blocTypeName(BlocType type) => switch (type) {
+  .bloc => 'Bloc',
+  .cubit => 'Cubit',
+};
```
PiotrRogulski: Use `.name` directly as an arg (no need to capitalize)

PiotrRogulski (approved):
One last comment, other than that LGTM <3

PR #546, opened 2026-07-15
Add `avoid_direct_collection_equality_checks` lint

cupofme:
Comparing collections with ==/!= in Dart checks identity, not contents — so [1, 2] == [1, 2] is false. It's an easy mistake to make and a painful one to debug, since the code reads like it should work.

This adds a lint that flags direct ==/!= comparisons between List, Set, or Map values (and their subtypes) when both sides are the same kind of collection.

Each warning comes with two quick fixes:
- `listEquals(a, b) `/ `setEquals` / `mapEquals` from package:flutter/foundation.dart
- const `ListEquality<int>().equals(a, b)` / `SetEquality` / `MapEquality` from package:collection

The messages and fix names adapt to the collection type (e.g. it suggests listEquals for a list, mapEquals for a map), and the package:collection fix fills in the type argument so you don't get an inference warning afterwards.

https://github.com/user-attachments/assets/fdf2ff56-22c6-4734-bc27-f2496d4f36d9


packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
```diff
@@ -0,0 +1,253 @@
...
+        ? leftType.asInstanceOf(collectionElement)?.typeArguments ?? const []
+        : const <DartType>[];
+
+    final negate = binary.operator.type == TokenType.BANG_EQ;
+    final left = binary.leftOperand.toSource();
+    final right = binary.rightOperand.toSource();
+
+    await builder.addDartFileEdit(file, (builder) {
+      builder
+        ..importLibraryElement(.parse('package:collection/collection.dart'))
```
PiotrRogulski: This fix should only be offered if the package containing the analyzed file depends on `package:collection`

packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
```diff
@@ -0,0 +1,253 @@
...
+        leftKind.collectionClass,
+      ],
+    );
+  }
+}
+
+/// The kind of a core collection type, used to pick the matching equality
+/// helper for the quick fixes.
+enum CollectionKind {
+  list('List', flutterFn: 'listEquals', collectionClass: 'ListEquality'),
```
PiotrRogulski: Don't abbreviate

packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
```diff
@@ -0,0 +1,253 @@
...
+
+/// Displays a warning when collections are compared directly with `==` or `!=`.
+class AvoidDirectCollectionEqualityChecks extends AnalysisRule {
+  AvoidDirectCollectionEqualityChecks()
+    : super(name: code.lowerCaseName, description: code.problemMessage);
+
+  static const code = LintCode(
+    'avoid_direct_collection_equality_checks',
+    'Avoid comparing {0}s directly with `==` or `!=`. This compares identity, not contents.',
+    correctionMessage: 'Use `{1}` or `const {2}().equals` instead.',
```
PiotrRogulski: Sometimes comparing identity is intended – we could also offer `identical()` as one of the fixes

packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
cupofme: Fixed

packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
cupofme: Good point. Added `identical` fix

packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
cupofme: God point, added this check. 

Current implementation only checks for a direct dependency in the `pubspec.yaml` of the package. There is still an edge case when other direct dependency can itself export `collection` package, but I don't think it's worth covering.

packages/leancode_lint/README.md
PiotrRogulski, line 1: Add `identical()` as "good"

packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
```diff
@@ -0,0 +1,328 @@
...
+CollectionKind? collectionKind(DartType? type) {
+  if (type == null) {
+    return null;
+  }
+
+  // `Map` is checked first because it is not an `Iterable`, while `Set` and
+  // `List` both are.
+  const map = TypeChecker.fromName('Map', packageName: 'dart:core');
+  const set = TypeChecker.fromName('Set', packageName: 'dart:core');
+  const list = TypeChecker.fromName('List', packageName: 'dart:core');
```
PiotrRogulski: You could use `isDartCore{type}` getters (e.g. https://pub.dev/documentation/analyzer/latest/dart_element_type/DartType/isDartCoreMap.html)

packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
```diff
@@ -0,0 +1,328 @@
...
+      return;
+    }
+
+    final negate = binary.operator.type == TokenType.BANG_EQ;
+    final left = binary.leftOperand.toSource();
+    final right = binary.rightOperand.toSource();
+
+    await builder.addDartFileEdit(file, (builder) {
+      builder
+        ..importLibraryElement(.parse('package:flutter/foundation.dart'))
```
PiotrRogulski: Similar to `package:collection` – this should only be available when Flutter is available

packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
PiotrRogulski: I don't think we should be checking pubspec.yaml. Get the package config instead

cupofme: Updated to something similar to what what [depend_on_referenced_packages ](https://github.com/dart-lang/sdk/blob/main/pkg/linter/lib/src/rules/pub/depend_on_referenced_packages.dart) is doing

packages/leancode_lint/README.md
cupofme: done

packages/leancode_lint/example/default/lib/avoid_direct_collection_equality_checks.dart
PiotrRogulski, line 1: I'm not sure if we need this file 🤔 The readme already contains some examples of good<->bad, so do tests — plus, we don't have such files for other lints. (I'm not against it per se, it's just that it's out of place for now — maybe we can add similar examples for other lints as well). Besides, doesn't this pollute the IDE analysis warnings? 🙈

packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
```diff
@@ -0,0 +1,289 @@
...
+/// Also matches subtypes of `List`, `Set` and `Map`.
+CollectionKind? collectionKind(DartType? type) {
+  if (type is! InterfaceType) {
+    return null;
+  }
+
+  final types = [type, ...type.allSupertypes];
+
+  // `Map` is checked first because it is not an `Iterable`, while `Set` and
+  // `List` both are.
```
PiotrRogulski: Does it matter? A map or its subclass won't pass the `isDartCoreSet` or `isDartCoreList` checks

packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
```diff
@@ -0,0 +1,289 @@
...
+    "Replace with '{0}'",
+  );
+
+  @override
+  List<String>? get fixArguments {
+    final binary = _targetBinary(node);
+    final kind = binary == null
+        ? null
+        : collectionKind(binary.leftOperand.staticType);
+    return [(kind ?? CollectionKind.list).flutterFunction];
```
PiotrRogulski: Why fallback to `.list`? I think that, at this point, there will always be a binary expression, so `!` should be "safe"

packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
```diff
@@ -0,0 +1,289 @@
...
+      return;
+    }
+
+    // Resolve the collection's type arguments (e.g. `int` for `List<int>`) so
+    // the generated constructor is `const ListEquality<int>()` rather than a
+    // raw `const ListEquality()`, which fails type inference.
+    final collectionElement = switch (kind) {
+      CollectionKind.list => typeProvider.listElement,
+      CollectionKind.set => typeProvider.setElement,
+      CollectionKind.map => typeProvider.mapElement,
```
PiotrRogulski:
```suggestion
      .list => typeProvider.listElement,
      .set => typeProvider.setElement,
      .map => typeProvider.mapElement,
```

packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
```diff
@@ -0,0 +1,289 @@
...
+          if (typeArguments.isNotEmpty) {
+            builder.write('<');
+            for (var i = 0; i < typeArguments.length; i++) {
+              if (i > 0) {
+                builder.write(', ');
+              }
+              builder.writeType(typeArguments[i], shouldWriteDynamic: true);
+            }
+            builder.write('>');
+          }
```
PiotrRogulski, line 239: There is a `builder.writeTypeParameters`; maybe it'll work 🤔

packages/leancode_lint/test/test_cases/avoid_direct_collection_equality_checks_test.dart
PiotrRogulski, line 1: Maybe we can also have test cases for `to{Type}()` methods:

```dart
List get valA;
Iterable get valB;

return valA == valB.toList();
//     ^^^^^^^^^^^^^^^^^^^^^
```

packages/leancode_lint/example/default/lib/avoid_direct_collection_equality_checks.dart
cupofme: This was the file I used to manually test the rule that I accidentally committed. Removed now

packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
cupofme: No, it doesn't. It's a pure AI slop comment that I missed. Thanks for noticing

packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
cupofme: done

packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
cupofme: From what I understand `writeTypeParameters` is supposed to be used for generating new generic declarations, not for supplying already-resolved type arguments to a constructor call. But I've replaced a loop with `writeTypes`

packages/leancode_lint/lib/src/lints/avoid_direct_collection_equality_checks.dart
```diff
@@ -0,0 +1,286 @@
...
+  FixKind get fixKind => const .new(
+    'leancode_lint.fix.replaceWithFlutterFoundationEquals',
+    DartFixKindPriority.standard,
+    "Replace with '{0}'",
+  );
+
+  @override
+  List<String>? get fixArguments {
+    final binary = _targetBinary(node);
+    final kind = binary == null
```
PiotrRogulski: I'd do `binary!` as well in these `fixArguments` getters

PiotrRogulski (approved):

PR #547, opened 2026-07-16
Add `avoid_context_read_in_build` lint

cupofme:
This PR adds a lint for easy to overlook bug: reading a value with `context.read` inside `build` method. `read` takes a one-time snapshot and never re-subscribes, so the UI quietly goes stale — reasonable-looking code that's easy to miss in review and annoying to debug later.

Now the analyzer flags it the moment it's written and offers a one-click fix to swap `read` for `watch`. It stays quiet on all the legitimate uses — calling a method, adding an event, grabbing a bloc reference, or reading inside a callback.

```dart
// BAD — reads a snapshot, UI won't rebuild when it changes
final count = context.read<CounterCubit>().state;
return Text('$count');

// GOOD — watch keeps the widget in sync
final count = context.watch<CounterCubit>().state;
return Text('$count');

// GOOD — reads that aren't build data stay quiet
onPressed: () => context.read<CounterCubit>().increment(),
```

https://github.com/user-attachments/assets/29799053-fe44-4b26-ae39-c84a07d81c1f

### DIFF
--- a/packages/leancode_lint/lib/src/lints/avoid_context_read_in_build.dart
+++ b/packages/leancode_lint/lib/src/lints/avoid_context_read_in_build.dart
@@ -5,8 +5,6 @@ import 'package:analyzer/analysis_rule/rule_context.dart';
 import 'package:analyzer/analysis_rule/rule_visitor_registry.dart';
 import 'package:analyzer/dart/ast/ast.dart';
 import 'package:analyzer/dart/ast/visitor.dart';
-import 'package:analyzer/dart/element/element.dart';
-import 'package:analyzer/dart/element/type.dart';
 import 'package:analyzer/error/error.dart';
 import 'package:analyzer_plugin/utilities/change_builder/change_builder_core.dart';
 import 'package:analyzer_plugin/utilities/fixes/fixes.dart';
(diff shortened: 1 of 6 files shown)


packages/leancode_lint/test/test_cases/avoid_context_read_in_build_test.dart
```diff
@@ -0,0 +1,194 @@
...
+    return const SizedBox();'''),
+    );
+  }
+
+  Future<void> test_tracedMethodReceiver_ok() async {
+    await assertNoDiagnostics(
+      _widget('''
+    final c = context.read<MyCubit>();
+    c.doThing();
+    return const SizedBox();'''),
```
PiotrRogulski: I'm not sure if this is OK – it's still top-level read in build. I feel like we should flag ***all*** reads, no matter what the purpose is 🤔

Invoking a method is still ***using*** the value

cupofme: Now all build-methods reads are flagged except for ones that are used in callbacks

packages/leancode_lint/lib/src/lints/avoid_context_read_in_build.dart
```diff
@@ -0,0 +1,141 @@
...
+import 'package:analyzer_plugin/utilities/fixes/fixes.dart';
+import 'package:analyzer_plugin/utilities/range_factory.dart';
+import 'package:leancode_lint/src/helpers.dart';
+import 'package:leancode_lint/src/type_checker.dart';
+
+/// Warns when `context.read` is called during `build`.
+///
+/// `read` grabs a value once and never re-subscribes, so using its result to
+/// render leaves the UI stale when the value changes — `watch` (or a
+/// `BlocBuilder`/`BlocSelector`) is what's actually wanted.
```
PiotrRogulski: select() is also valid (just like BlocSelector). We probably can't suggest it in the fix (because it would require rewriting more code), but we should mention it in docs and messages

PiotrRogulski (commented):
One last thing 🙏

packages/leancode_lint/lib/src/lints/avoid_context_read_in_build.dart
cupofme: done

PiotrRogulski (approved):

PR #552, opened 2026-08-03
Add pana-score CI step to all package test workflows

PiotrRogulski:
Runs leancodepl/mobile-tools pana-score action after each package's test job to surface pub.dev-style scoring as a commit status and step summary, without failing CI on a low score.


mateusz-pietras (approved):
LGTM

PR #553, opened 2026-08-03
[leancode_debug_page] Keep text legible on the debug page's own colors

Komoszek:
The log tiles, the log details app bar and the entry button paint themselves with colors the app's theme knows nothing about, but left the text and icons on top of them to the theme's foregrounds. Derive the foreground from the painted color instead.

`ThemeData.estimateBrightnessForColor` won't do it: it reads mid tones like `Colors.green` as dark, which would put white on them at 2.8:1. The threshold here is the luminance where black and white contrast equally, so every status color gets 5.7:1 or better.

### DIFF
--- a/packages/leancode_debug_page/lib/src/ui/status_color_extension.dart
+++ b/packages/leancode_debug_page/lib/src/ui/status_color_extension.dart
@@ -1,0 +1,16 @@
+import 'dart:ui';
+
+import 'package:flutter/material.dart' show Colors, ThemeData;
+
+extension StatusColorExtension on Color {
+  /// A foreground legible on a surface painted with this color.
+  ///
+  /// The debug page paints tiles and app bars with its own status colors, which
+  /// the app's theme knows nothing about, so its `onSurface` cannot be trusted
+  /// on them. [ThemeData.estimateBrightnessForColor] answers a different
+  /// question - it reads mid tones like [Colors.green] as dark and would put
+  /// white on them at 2.8:1 - so compare against the luminance where black and
+  /// white contrast equally, `sqrt(1.05 * 0.05) - 0.05` per WCAG.
+  Color get onStatusColor =>
+      computeLuminance() > 0.179 ? Colors.black : Colors.white;
+}
--- a/packages/leancode_debug_page/lib/src/ui/status_color_extension.dart
+++ b/packages/leancode_debug_page/lib/src/ui/status_color_extension.dart
@@ -1,4 +1,14 @@
+import 'dart:math';
+
 import 'package:flutter/material.dart';
+
+/// The luminance at which black and white contrast equally.
+///
+/// WCAG rates the contrast of two luminances as
+/// `(lighter + 0.05) / (darker + 0.05)`, so this solves
+/// `1.05 / (l + 0.05) == (l + 0.05) / 0.05` - the point where a foreground is
+/// no better off white than black.
+final _equalContrastLuminance = sqrt(1.05 * 0.05) - 0.05;
 
 extension StatusColorExtension on Color {
   /// A foreground legible on a surface painted with this color.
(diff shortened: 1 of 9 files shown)


packages/leancode_debug_page/lib/src/ui/status_color_extension.dart
```diff
@@ -0,0 +1,16 @@
...
+
+extension StatusColorExtension on Color {
+  /// A foreground legible on a surface painted with this color.
+  ///
+  /// The debug page paints tiles and app bars with its own status colors, which
+  /// the app's theme knows nothing about, so its `onSurface` cannot be trusted
+  /// on them. [ThemeData.estimateBrightnessForColor] answers a different
+  /// question - it reads mid tones like [Colors.green] as dark and would put
+  /// white on them at 2.8:1 - so compare against the luminance where black and
+  /// white contrast equally, `sqrt(1.05 * 0.05) - 0.05` per WCAG.
```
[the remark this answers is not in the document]
Komoszek: `unused_import` couldn't have fired here — `Colors.black`/`Colors.white` come from that same import and are used in code, so the import as a whole was always live. The analyzer also resolves doc comment references, so a doc-only `[ThemeData]` doesn't count as unused either.

The `show` combinator was still worth dropping: it was the only one in the repo, and every other file imports `material.dart` plainly. Done in 406ccf8, with the dartdoc links kept rather than flattened to plain text.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/leancode_debug_page/lib/src/ui/status_color_extension.dart
```diff
@@ -0,0 +1,14 @@
...
+  Color get onStatusColor =>
+      computeLuminance() > 0.179 ? Colors.black : Colors.white;
```
PiotrRogulski: Let's use this formula in code and calculate the threshold (into a private const/final here) – the WCAG doc can be on that field)

packages/leancode_debug_page/test/ui/log_tile_contrast_test.dart
```diff
@@ -0,0 +1,116 @@
...
+    colorScheme: ColorScheme.fromSeed(
+      seedColor: Colors.blue,
+      brightness: brightness,
+    ).copyWith(onSurface: const Color(0xFF808080)),
+  );
+
+  double contrastRatio(Color a, Color b) {
+    final luminances = [a.computeLuminance(), b.computeLuminance()];
+
+    return (luminances.reduce(max) + 0.05) / (luminances.reduce(min) + 0.05);
```
PiotrRogulski: Using `.reduce` for a two-element list is overkill lol


```suggestion
    final luminanceA = a.computeLuminance();
    final luminanceB = b.computeLuminance();

    return (max(luminanceA, luminanceB) + 0.05) / (min(luminanceA, luminanceB) + 0.05);
```

^ with shorter var names will look nicer

packages/leancode_debug_page/lib/src/ui/status_color_extension.dart
Komoszek: Done in 0e85f4f — the threshold is now a private `final _equalContrastLuminance = sqrt(1.05 * 0.05) - 0.05;` with the WCAG derivation as its doc, and the getter's doc keeps only the `estimateBrightnessForColor` rationale. `final` rather than `const` since `sqrt` isn't const-evaluable.

Solving it in code moves the threshold from 0.179 to 0.17913, which doesn't flip any status color — the closest is `green.shade800` at L=0.155. Existing contrast tests still cover it.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/leancode_debug_page/test/ui/log_tile_contrast_test.dart
Komoszek: Fair — done in 92b71d7. Went with `lumA`/`lumB`, which keeps the return on one line:

```dart
final lumA = a.computeLuminance();
final lumB = b.computeLuminance();

return (max(lumA, lumB) + 0.05) / (min(lumA, lumB) + 0.05);
```

---
_Generated by [Claude Code](https://claude.ai/code)_

PiotrRogulski (approved):

PR #556, opened 2026-08-07
Allow analyzer 14

PiotrRogulski:
This PR ***broadens*** the allowed range for `analyzer` to still support versions 13.x — there have been no breaking changes, and many packages don't support analyzer 14 yet.

This also wins us back points on pub.dev (from the `Support up-to-date dependencies` category) – and puts us back at full 160/160.

> [!NOTE]
> I already bumped the package version to 26.1.0 instead of adding this change to an `Unreleased` section — I'll release this right after merging


mateusz-pietras (approved):
LGTM

PR #559, opened 2026-08-09
[leancode_force_update] Upgrade contracts dependencies

PiotrRogulski:
Upgrades the contracts toolchain in `leancode_force_update` and regenerates the contracts.

## Changes

- `leancode_contracts` `^0.9.0` → `^0.10.0`
- `leancode_contracts_generator` `^0.18.0` → `^0.19.0`
- `dotnet-contracts-generate` (`.config/dotnet-tools.json`) `4.0.0` → `4.1.0`
- `leancode_lint` `^17.0.0` → `^19.0.0` — forced by the above: lint 17/18 pin `analyzer ^7.3.0` while the new generator needs `analyzer >=8.1.1`, so version solving failed otherwise. 19.0.x is the lowest version that resolves while keeping the package's `sdk: >=3.10.0` floor (lint ≥20 would require bumping the SDK/Flutter constraints and the CI matrix).
- Dropped the `json_serializable` dev dependency — generator v0.19.0 generates serialization with its own `contracts_serializable` builder, so the `json_serializable` builder had become a no-op here. It remains in the lockfile as a transitive dependency of the generator. `build_runner` stays, since it drives the new builder.
- Version bumped to `1.0.4-alpha` (`1.0.3-alpha` is already published), with a CHANGELOG entry.

`Leancode.ForceUpdate.Contracts` is left at `10.0.2860` — already the newest published version.

## Regenerated output

`contracts.dart` is unchanged. The only change in `contracts.g.dart` is the generator header:

```diff
-// JsonSerializableGenerator
+// ContractsSerializableGenerator
```

The `fromJson`/`toJson` bodies are byte-identical, both across the generator upgrade and across the `dotnet-contracts-generate` bump.

## Verification

Ran locally against Flutter 3.44.9 stable and .NET SDK 8:

- Confirmed the pre-upgrade toolchain reproduced the committed generated files byte-for-byte before changing anything, to validate the setup.
- `flutter analyze` — no issues
- `flutter test` — 8/8 passing
- `flutter pub publish --dry-run` — clean apart from the usual "not an incremental update" hint that the `-alpha` versioning scheme has always produced

---
_Generated by [Claude Code](https://claude.ai/code/session_01CwSEMS3JeqqcupWaZcS18h)_

### DIFF
--- a/packages/leancode_force_update/.config/dotnet-tools.json
+++ b/packages/leancode_force_update/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "dotnet-contracts-generate": {
-      "version": "4.0.0",
+      "version": "4.1.0",
       "commands": [
         "dotnet-contracts-generate"
       ],
--- a/packages/leancode_force_update/CHANGELOG.md
+++ b/packages/leancode_force_update/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 1.0.4-alpha
+
+- Upgrade `leancode_contracts` dependency to `^0.10.0`
+- Upgrade `leancode_contracts_generator` dev dependency to `^0.19.0`
+- Upgrade `leancode_lint` dev dependency to `^19.0.0`
+- Upgrade `dotnet-contracts-generate` to `4.1.0`
+- Regenerate contracts
+
 ## 1.0.3-alpha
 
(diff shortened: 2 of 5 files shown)


PiotrRogulski:
pana reports 150/160 because we don't support `in_app_update 5.0.0` — I'll upgrade it in a separate PR

Komoszek (approved):

PR #563, opened 2026-08-10
[leancode_analytics_posthog] Send one $screen event per navigation

zoskar:
⚠️  - 🤖 AI fix for sth that came out in new project made from template

## Problem

`LeanAnalyticsPostHogObserver` sends a `$screen` for the `LeanAnalyticsRoute` id, then calls `super`,
which makes `PosthogObserver` send a second one named after `RouteSettings.name`.

One navigation, two events, two naming schemes — in an auto_route app that's `loginPage` and
`LoginRoute` for the same push. Screen view counts come out doubled and split across both names.

It only surfaces when routes carry a non-empty `RouteSettings.name`; `PosthogObserver` skips the
rest, which is why the package can look fine in projects that don't name their routes.

## Fix

Pass a `nameExtractor` that returns `null`, so the inherited screen view is suppressed while
`PosthogObserver`'s own bookkeeping — survey context, route subscriptions — still runs.

Routes that don't implement `LeanAnalyticsRoute` stop being tracked, which is what the class already
documents.

### DIFF
--- a/packages/leancode_analytics_posthog/CHANGELOG.md
+++ b/packages/leancode_analytics_posthog/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.1.1
+
+- Fix `LeanAnalyticsPostHogObserver` sending two `$screen` events per navigation
+
 ## 0.1.0
 
 - Initial release with PostHog Analytics implementation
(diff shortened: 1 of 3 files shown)


PiotrRogulski:
@zoskar I think a more "complete" fix would be to implement support for parameters in the `posthog_flutter` package itself 😛 (e.g., there could be a `parametersExtractor` next to the exxisting `nameExtractor`)

PR #570, opened 2026-09-10
Add jaspr_class_scope package for CSS classname scoping

Komoszek:
## Problem

Jaspr collects every `@css` getter in the app into one stylesheet for the page, and class names in it
are plain strings. Two components that both call something `title` style each other, a rename in one
file repaints another, and a selector can drift from the `classes:` attribute it was written for.
Nothing on pub.dev scopes class names for Jaspr; the advanced_forms landing page needed it.

## What this adds

`jaspr_class_scope` (no dependencies) holds the vocabulary: a `ClassName` is written once and asked
for both spellings — `name` for `classes:`, `selector` for the rule — so the two cannot drift.
`ClassName.shared` renders a name as written, for the classes a hand-written stylesheet or a bit of
JavaScript looks up.

`jaspr_class_scope_builder` (a dev dependency) gives each component its own namespace:

```dart
part 'hero.scopes.dart';

@scopedCss
class Hero extends StatelessComponent {
  static const _class = _$HeroScope;
  static final _grid = _class('grid');
}
```

`_grid` renders as `grid-<suffix>`, the suffix being the md5 of `package|path#Component` in six
base-36 digits. Build time, because `Type.hashCode` differs between the VM and JS, a type's name is
gone after minification, and a scope has to be `const` to be spelled in a `@css` getter.

Two components can still land on the same suffix, rarely. A second build phase catches that and fails
the build with both names. It reads a manifest each package writes into the build cache, so it covers the
whole package graph — a component of a dependency is checked against yours.

## Test plan

- [x] `dart test` — 6 in `jaspr_class_scope`, 14 in the builder, including real colliding pairs found
      by searching the hash, within one package and across two
- [x] `dart analyze --fatal-infos` and `dart format` clean; compiles against analyzer 12 and 14, since
      `jaspr_builder` 0.23 pins `analyzer: ^12.1.0`
- [x] Drives the advanced_forms landing page (leancodepl/advanced_forms#80): 21 scopes over 11
      components, 67 scoped class names in the built page, and the names shared with `landing.js` and
      the docs' stylesheet unchanged

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XirTyFcvXRo3Hs9PBUnt7r


packages/jaspr_class_scope/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2023 LeanCode Sp. z o.o.
```
PiotrRogulski: 2026? (The other license file as well)

packages/jaspr_class_scope/example/hero.dart
```diff
@@ -0,0 +1,23 @@
...
+
+@scopedCss
+class Hero {
+  static const _class = _$heroScope;
+
+  static final _grid = _class('grid');
+  static final _title = _class('title');
+
+  /// What a `@css` getter would style.
+  static String get styles => '${_grid.selector} > ${_title.selector}';
```
PiotrRogulski, line 16: Maybe we can ship a way to combine selectors like this as well (not only `+`)?

PiotrRogulski: Also, as an example, I feel like this getter might be a bit confusing? The `styles` getter in Jaspr contains entire CSS styles, but this one is just a selector

packages/jaspr_class_scope/lib/src/class_name.dart
```diff
@@ -0,0 +1,56 @@
...
+  const ClassName.shared(String name)
+    : _local = name,
+      _scope = null,
+      _and = null;
+
+  /// The class [local] of [scope]'s component.
+  const ClassName.scoped(String local, ClassScope scope)
+    : _local = local,
+      _scope = scope,
+      _and = null;
```
PiotrRogulski:
```suggestion
  const ClassName.shared(this._local)
    : _scope = null,
      _and = null;

  /// The class [local] of [scope]'s component.
  const ClassName.scoped(this._local, this._scope)
    : _and = null;
```

packages/jaspr_class_scope_builder/lib/src/scoped_components.dart
```diff
@@ -0,0 +1,67 @@
+import 'package:analyzer/dart/analysis/utilities.dart';
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:build/build.dart';
+import 'package:jaspr_class_scope_builder/src/suffix.dart';
+
+/// The extension of the part file the scopes are written to.
+const scopesExtension = '.scopes.dart';
+
+/// Where a package lists its scopes for the check.
+const manifestPath = 'lib/jaspr_class_scope.scopes.json';
```
PiotrRogulski: Should we actually write this file into `lib`, and not some temp (dart-build-specific) directory? `flutter_gen` does this

packages/jaspr_class_scope/example/hero.dart
Komoszek: > Maybe we can ship a way to combine selectors like this as well (not only +)?

I would rather not do it. We would have to create an alternative syntax for the all CSS selectors for not a lot (if even any) benefits

> The `styles` getter in Jaspr contains entire CSS styles, but this one is just a selector

I don't think this is a problem. This is just an example

packages/jaspr_class_scope_builder/lib/src/scoped_components.dart
```diff
@@ -0,0 +1,69 @@
...
+String scopeSourceOf(AssetId asset, String component) =>
+    '${asset.package}|${asset.path}#$component';
+
+/// How a clash names [component] in [asset].
+String scopeOwnerOf(AssetId asset, String component) =>
+    '$component (${asset.package}|${asset.path})';
+
+/// The scope of [component] in [asset], as the part file spells it.
+String renderScope(AssetId asset, String component) {
+  final constant = '_\$${component[0].toLowerCase()}${component.substring(1)}';
```
PiotrRogulski: Maybe we could keep the component name as-is in the generated scope name?

packages/jaspr_class_scope_builder/lib/src/scoped_components.dart
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// Matched by the name the annotation is spelled with: resolving it costs a
+// great deal more, and the class name is all this needs.
+bool _isScopedCss(ClassDeclaration declaration) =>
+    declaration.metadata.any((annotation) {
+      final name = annotation.name.name.split('.').last;
+
+      return name == 'scopedCss' || name == 'ScopedCss';
+    });
```
PiotrRogulski: Let's also check the library where the annotation comes from

packages/jaspr_class_scope_builder/build.yaml
```diff
@@ -0,0 +1,29 @@
+# Keep these extensions in step with `lib/src/`; build_runner reads this file
+# without loading Dart.
+builders:
+  class_scope:
+    import: 'package:jaspr_class_scope_builder/jaspr_class_scope_builder.dart'
+    builder_factories: ['classScopeBuilder']
+    build_extensions: {'.dart': ['.scopes.dart']}
```
PiotrRogulski: Why not use yaml "block" syntax, with indentation? (in this entire file)

packages/jaspr_class_scope_builder/lib/src/class_scope_check_builder.dart
```diff
@@ -0,0 +1,64 @@
...
+    r'$package$': ['jaspr_class_scope.check'],
+  };
+
+  @override
+  Future<void> build(BuildStep buildStep) async {
+    // Sorted, so that a clash is reported the same way on every machine.
+    final packages = SplayTreeSet<String>.of([
+      buildStep.inputId.package,
+      ...(await buildStep.packageConfig).packages.map((it) => it.name),
+    ]);
```
PiotrRogulski: Why not just


```suggestion
    final packages = [
      buildStep.inputId.package,
      ...(await buildStep.packageConfig).packages.map((it) => it.name),
    ]..sort();
```

Sorting the list and creating a splay tree would take the same time, but iterating on a list should be faster

packages/jaspr_class_scope_builder/lib/src/class_scope_check_builder.dart
```diff
@@ -0,0 +1,64 @@
...
+        continue;
+      }
+
+      final listed =
+          (jsonDecode(await buildStep.readAsString(manifest)) as List<Object?>)
+              .cast<Map<String, Object?>>();
+
+      for (final scope in listed) {
+        final suffix = scope['suffix']! as String;
+        final owner = scope['owner']! as String;
```
PiotrRogulski:
```suggestion
      for (final {'suffix': suffix as String, 'owner': owner as String} in listed) {
```

packages/jaspr_class_scope_builder/lib/src/scoped_components.dart
```diff
@@ -0,0 +1,79 @@
...
+// What says the annotation is this package's: the file asks for the part file
+// only this builder writes. An import cannot say it, since a project may
+// re-export the annotation, and resolving it would tie the builder to one
+// version of the analyzer.
+bool _declaresScopes(CompilationUnit unit) => unit.directives
+    .whereType<PartDirective>()
+    .any((it) => it.uri.stringValue?.endsWith(scopesExtension) ?? false);
+
+// Matched by the name the annotation is spelled with, which the part directive
+// above has already tied to this package.
```
PiotrRogulski: > which the part directive above has already tied to this package.

What does this mean? How does a part directive ensure that we only see correct annotations?

packages/jaspr_class_scope/LICENSE
Komoszek: Right — both files were copied from an older package and kept its year. `Copyright 2023` → `Copyright 2026` in `jaspr_class_scope` and `jaspr_class_scope_builder` (6e0c688).

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/jaspr_class_scope_builder/lib/src/scoped_components.dart
Komoszek: Done in 6e0c688 — the constant is `_$HeroScope` now.

It was worse than lowercasing one letter: `CTAButton` came out as `_$cTAButtonScope`. Keeping the name as written also matches what the rest of the ecosystem generates (`_$PersonFromJson`, `_$$PersonImpl`).

Checked that a capital right after `_$` doesn't trip `constant_identifier_names` under `leancode_lint` — the generated part file sits in the consumer's `lib/` and is analyzed with their rules. The suffixes are unchanged, so no rendered class name moved.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/jaspr_class_scope_builder/lib/src/scoped_components.dart
Komoszek: The risk is real, but neither obvious way of checking works here:

- **The imports.** The file carrying `@scopedCss` often doesn't import this package at all. The advanced_forms landing re-exports `jaspr_class_scope` from its own `styles.dart`, and every component imports that one file — an import check would silently stop generating scopes for the whole site.
- **Resolution** (`buildStep.resolver`). Answers it exactly, and costs two things: it pins the builder to one analyzer major, while `jaspr_builder` 0.23 still requires `analyzer: ^12.1.0` and we compile against both 12 and 14; and the manifest builder runs with `auto_apply: all_packages`, so it would mean resolving every library in the package graph to find annotations in a handful of files.

So 6e0c688 keys off the `part` directive instead: a class is this package's to scope only if its file asks for the `.scopes.dart` part file that only this builder writes. Someone else's `@scopedCss` would have to also request our part file to be picked up, which is not an accident anyone has by mistake. It stays syntactic, so it costs nothing and holds across analyzer versions.

It also fixes a real bug on the way: `@scopedCss` without the `part` directive used to generate a file with an orphan `part of` pointing at a library that never declared it. Now it generates nothing, and the undefined `_$…Scope` in your own file is what tells you the directive is missing. Covered by `writes nothing for a file that declares no part`.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/jaspr_class_scope_builder/build.yaml
Komoszek: No reason beyond how it was written — switched to block syntax through the whole file in 6e0c688, and checked with a YAML parser that the three builders come out identical.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/jaspr_class_scope_builder/lib/src/class_scope_check_builder.dart
Komoszek: Taken, with the set kept for one job only — deduplication. `packageConfig` already lists the package the check runs in, so with a plain list its manifest is read twice and the second pass reports it as clashing with itself. Applying the suggestion verbatim fails the check's own test:

```
Bad state: Card (site|lib/card.dart) and Card (site|lib/card.dart) both scope to the same suffix.
```

So c642517 builds the set, drops to a list to iterate, and says why in a comment:

```dart
final packages = {
  buildStep.inputId.package,
  ...(await buildStep.packageConfig).packages.map((it) => it.name),
};

// Sorted, so that a clash is reported the same way on every machine.
for (final package in packages.toList()..sort()) {
```

The `SplayTreeSet` was there to keep `dart format` from mangling a `..sort()` cascade on a collection literal; moving the sort into the `for` header solves that without it.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/jaspr_class_scope_builder/lib/src/class_scope_check_builder.dart
Komoszek: Applied in c642517. Worth noting for anyone copying it: the `as String` casts are what make it work — a declaration pattern (`'suffix': String suffix`) is refutable against `Object?` and the analyzer rejects it in a for-in with `pattern_type_mismatch_in_irrefutable_context`.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/jaspr_class_scope_builder/lib/src/scoped_components.dart
Komoszek: It doesn't *ensure* it, and the comment claimed more than it should. Reworded in c642517:

```dart
// Matched by the name the annotation is spelled with. The part directive above
// is what makes a name enough: another package's `@scopedCss` would have to sit
// in a file that also asks for this builder's part file.
```

The mechanism, in full: `scopedComponentsIn` now returns nothing unless the file declares `part '….scopes.dart'`. That part file is written by this builder and nobody else, so a file asking for it is asking for this generator. A same-named annotation from some other package is then harmless unless its author also wrote a part directive pointing at a file only we produce.

The reason it isn't the import or the resolved element is in [the other thread](https://github.com/leancodepl/flutter_corelibrary/pull/570#discussion_r4005730817) — short version: an import check breaks any project that re-exports the annotation through its own file (the advanced_forms landing does exactly that), and resolving it pins the builder to one analyzer major while `jaspr_builder` 0.23 still requires `analyzer: ^12.1.0`.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/jaspr_class_scope_builder/lib/src/scoped_components.dart
```diff
@@ -0,0 +1,80 @@
...
+// only this builder writes. An import cannot say it, since a project may
+// re-export the annotation, and resolving it would tie the builder to one
+// version of the analyzer.
+bool _declaresScopes(CompilationUnit unit) => unit.directives
+    .whereType<PartDirective>()
+    .any((it) => it.uri.stringValue?.endsWith(scopesExtension) ?? false);
+
+// Matched by the name the annotation is spelled with. The part directive above
+// is what makes a name enough: another package's `@scopedCss` would have to sit
+// in a file that also asks for this builder's part file.
```
PiotrRogulski: > another package's `@scopedCss` would have to sit in a file that also asks for this builder's part file.

Huh??? Why are you so against just checking the defining library???

packages/jaspr_class_scope_builder/lib/src/class_scope_check_builder.dart
PiotrRogulski: Ok, so if the package config always contains the current package as well, let's just not include `buildStep.inputId.package`? Then there won't be duplicates, no?

packages/jaspr_class_scope_builder/lib/src/scoped_components.dart
Komoszek: Fair — I was wrong about the cost, and I should have measured before arguing. It's the defining library now, in 6dd3dbd:

```dart
bool _isScopedCss(ClassElement component) =>
    component.metadata.annotations.any((annotation) {
      final uri = annotation.element?.library?.uri;

      return uri != null &&
          uri.scheme == 'package' &&
          uri.pathSegments.first == 'jaspr_class_scope';
    });
```

Both objections I raised turned out not to hold:

- **Analyzer versions.** The pieces this needs — `LibraryElement.classes`, `Element.library`, `Element.metadata`, `ElementAnnotation.element`, `LibraryElement.uri` — are spelled identically in analyzer 12.1.0 and 14.4.0. What moved in 14.3 was `ClassDeclaration.name`, an AST getter, which is why `_nameOf` reads tokens; the element model was never the problem. Verified by building the advanced_forms landing with `analyzer: 12.1.0` resolved.
- **Cost.** Resolution only happens for a file that names the annotation *and* declares the `.scopes.dart` part, so it is a handful of files per package. A clean build of that landing (76 packages in the graph, 21 scopes, no part files on disk) takes 39–40s against 46s before this change — i.e. no measurable difference.

One thing that did need care: on a first build the declared part file does not exist yet. That is a resolution error, not a syntax one, so `libraryFor(asset, allowSyntaxErrors: true)` still returns the library, and a clean build from zero part files works — that is what the landing run above proves.

Two tests now pin the behaviour: `skips a same-named annotation from another package` (a foreign `ScopedCss`/`scopedCss` pair, which the old name check accepted) and `finds the annotation through a re-export` (the landing's shape — its components import `styles.dart`, never this package).

The `part` directive stays, but only as a cheap filter before resolving and as the reason we never write a part file nobody declared. Ownership is the library check now.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/jaspr_class_scope_builder/README.md
```diff
@@ -0,0 +1,95 @@
...
+}
+```
+
+`dart run build_runner build` writes `hero.scopes.dart` next to it, holding
+`Hero`'s scope. `_grid` then renders as `grid` with a suffix belonging to that
+scope; the same class under `lib/marketing/` gets its own. Moving the file or
+renaming the class changes the scope. `CardGrid` gives `_$CardGridScope`. The
+annotation is matched by where it is declared, not by its name, so a same-named
+annotation from elsewhere is left alone and `@scopedCss` reached through your
+own re-export still counts. A file that declares no `part` produces nothing.
```
PiotrRogulski: > The annotation is matched by where it is declared, not by its name, so a same-named annotation from elsewhere is left alone

Yeah, I'm afraid that this is the expected and the only acceptable behavior — no need to explain this

packages/jaspr_class_scope_builder/lib/src/scopes_manifest_builder.dart
```diff
@@ -0,0 +1,54 @@
...
+        continue;
+      }
+
+      for (final component in await scopedComponentsIn(buildStep, asset)) {
+        scopes.add({
+          'suffix': classScopeSuffix(scopeSourceOf(asset, component)),
+          'owner': scopeOwnerOf(asset, component),
+        });
+      }
+    }
```
PiotrRogulski: Make this a list literal

packages/jaspr_class_scope_builder/lib/src/scoped_components.dart
```diff
@@ -0,0 +1,84 @@
...
+
+// The file asks for the part file this builder writes, which is both what
+// makes it ours to generate for and a reason not to resolve the rest.
+bool _declaresScopes(CompilationUnit unit) => unit.directives
+    .whereType<PartDirective>()
+    .any((it) => it.uri.stringValue?.endsWith(scopesExtension) ?? false);
+
+// Where the annotation is declared, not how it is spelled: a project may
+// re-export `scopedCss` under its own name, and another package may declare
+// something of the same name that is not ours.
```
PiotrRogulski: Please go over all the comments in this PR and keep only what's actually important. Stuff like this isn't

packages/jaspr_class_scope_builder/lib/src/class_scope_check_builder.dart
Komoszek: Yes — `BuildPackages.asPackageConfig` is built from every package in the build, the current one included. Dropped in 197e587; it's a plain sorted list now.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/jaspr_class_scope_builder/README.md
Komoszek: Removed in e885490.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/jaspr_class_scope_builder/lib/src/scopes_manifest_builder.dart
Komoszek: Done in e885490.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/jaspr_class_scope_builder/lib/src/scoped_components.dart
Komoszek: Done in e885490. Gone: this one, the one on `_declaresScopes`, the three-line preamble before the early exits, the `allowSyntaxErrors` note (and the flag itself — a missing part file is a resolution error, so `libraryFor` never needed it), the second paragraph of `ScopesManifestBuilder`'s doc, the `build.yaml` header, and the test comments that repeated their goldens. Kept the ones that carry a fact the code doesn't: why the manifest lives under `lib/`, why the packages are sorted, why `// dart format off`, why `all_packages`, and where the colliding test pairs come from.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/jaspr_class_scope_builder/lib/src/scoped_components.dart
Komoszek: Correction to the above: "what moved in 14.3 was `ClassDeclaration.name`" is wrong. It was deprecated in analyzer 10.0.0 and removed in 11.0.0, and `namePart.typeName` has been there since 10 — well below this package's 12.1.0 floor. So the token walk in `_nameOf` was never needed; it went away with the switch to the element model, but I stated a false reason for it here and in an earlier code comment.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/jaspr_class_scope_builder/lib/src/scoped_components.dart
Komoszek: Correction: `_$$PersonImpl` above was not checked. json_serializable does generate `_$PersonFromJson` (`decode_helper.dart:41`, `prefix => '_$${element.name.nonPrivate}'`), but freezed 4.0.1's only `_$…Impl` name is `_$PersonCopyWithImpl` (`copy_with.dart:347`); its concrete class carries no `Impl` at all. The point stands on json_serializable alone.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/jaspr_class_scope_builder/README.md
```diff
@@ -0,0 +1,93 @@
...
+class Hero extends StatelessComponent {
+  static const _class = _$HeroScope;
+
+  static final _grid = _class('grid');
+}
+```
+
+`dart run build_runner build` writes `hero.scopes.dart` next to it, holding
+`Hero`'s scope. `_grid` then renders as `grid` with a suffix belonging to that
+scope; the same class under `lib/marketing/` gets its own. Moving the file or
```
PiotrRogulski: > the same class under `lib/marketing/` gets its own

What's `marketing` doing here? I feel like this is just confusing. Please write a better and clearer readme xD

Komoszek: Rewritten in 9ae7b62. It now shows the generated `hero.scopes.dart`, says what the suffix is derived from, and drops the `marketing` aside.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/jaspr_class_scope_builder/README.md
```diff
@@ -0,0 +1,105 @@
...
+```dart
+part of 'hero.dart';
+
+const _$HeroScope = ClassScope('Hero', '7en7oj');
+```
+
+The suffix is derived from the package name, the file path and the class name.
+`_grid` renders as `grid-7en7oj`, and a `_class('grid')` in any other component
+renders with that component's suffix instead. Renaming or moving `Hero` changes
+it.
```
PiotrRogulski: > and a `_class('grid')` in any other component renders with that component's suffix instead

Yes, but I'd frame it in a different way — components that share the same name get different suffixes/classnames. What you wrote also includes components with entirely different names, which is vacuously true that they have different suffixes

Komoszek: Reframed in 999ecbf: "Two components that both name a class `grid`, or two files that both declare a `Hero`, get different suffixes."

---
_Generated by [Claude Code](https://claude.ai/code)_

PiotrRogulski: > Two components that both name a class `grid`, or two files that both declare a `Hero`, get different suffixes.

Not exactly??? The logic in this sentence is a bit backward. Sharing a class name does not affect the generated suffixes.

> The suffix is derived from the package name, the file path and the class name
Yes. However, this means that two components with the same name will get different suffixes. Just that, and only that

Komoszek: Right, the class name has nothing to do with it. a10cd8b: "…and two components with the same name get different suffixes."

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/jaspr_class_scope_builder/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+```dart
+part of 'hero.dart';
+
+const _$HeroScope = ClassScope('Hero', '7en7oj');
+```
+
+The suffix is derived from the package name, the file path and the class name,
+so `_grid` renders as `grid-7en7oj`, and two components with the same name get
+different suffixes. Renaming or moving `Hero` changes it.
```
PiotrRogulski: > changes it

What is "it"? Because if you mean the suffix, then yes. If the fact that they are different, then no

Komoszek: The suffix. a9032d9: "Renaming or moving `Hero` changes its suffix."

---
_Generated by [Claude Code](https://claude.ai/code)_

PiotrRogulski (approved):

